Issue #46, opened 2021-02-26
Debugger restart error if debug icon is pressed after a stop


rsora:
**Describe the bug**
Some times the debug button behavior triggers a failed launch of `openocd` probably due to a bug in the GUI code

**To Reproduce**
1. Open a sketch
2. Connect a debug enabled board (like an Arduino Zero)
2. Select **Sketch > Optimize for Debugging**
3. Upload Sketch
4. Click the Debug Button 
![image](https://user-images.githubusercontent.com/35916119/109290784-59781900-7828-11eb-8c0d-0cb60f643ee4.png)
5. Add a breakpoint in your sketch
6. Click "Continue" in the Debugger control bar
![image](https://user-images.githubusercontent.com/35916119/109291054-c12e6400-7828-11eb-9000-8cc5ee2be0a3.png)
7. Click "Stop" in the Debugger control bar
8. Click the Debug button again and wait
9. See the error (also reported in the screenshot below)
```
Failed to launch OpenOCD GDB Server: Error: spawn /home/rsora/.arduino15/packages/arduino/tools/openocd/0.10.0-arduino7/bin/openocd ENOENT
```
Please note that the last 3 steps may need to be performed a few times before seeing the failure.
Reproducibility is not consistent

**Screenshots**
![image](https://user-images.githubusercontent.com/35916119/109290271-aefff600-7827-11eb-91af-42b36b233480.png)

**Desktop (please complete the following information):**
- OS: we were able to reporduce the issue on Linux and MacOS and Windows
- Version: 2.0.0-beta.3-snapshot.9fb8c4b
- Date: 2021-02-25T15:29:51.087Z
- CLI Version: 0.16.1 alpha [76f55490]

PR #50, opened 2021-03-01
Fixed workspace variable resolver. Can start debug session multiple times from the toolbar.

kittaakos:
Fall back to the current sketch, if `currentWidget` points to a file
outside of the workspace.

Closes: arduino/arduino-ide#46

How to test:
 - You should be able to start/stop as many consecutive debug sessions from the toolbar as you pleased.


per1234 (approved):
I am able to reproduce https://github.com/arduino/arduino-ide/issues/46 when using 2.0.0-beta3. The issue does not occur when using the test build from this PR.

PR #168, opened 2021-03-03
ATL-972: Moved the `./theia/launch.json` config into a temp folder.

kittaakos:
What it does:
 - This PR gets rid of the `sketch_folder/.theia/launch.json` and moves it to a temp folder.

Steps to reproduce:
 - Create a sketch. Do not save it yet.
 - Select a board that can debug.
 - Compile and start the debug from the toolbar.
 - Debug view open, you should see a config entry `Arduino`. You can open the config file in an editor. Hover over the editor tab of `launch.json`; it will point to some temp folder (`arduino-ide2-`).
 - You should be able to re-launch the debug from the Debug view and toolbar as many times as you want.
 - Save the sketch. (**Note:** the `Arduino` config is missing from the debug view! We need a solution for this in the future.)
 - Compile and start the debugging. You should be able to restart the debug session from both the view and the toolbar.
 - Stop and start the IDE. Since the sketch was saved previously, you should see the `Arduino` debug config in the Debug view, and you can start the debug session from there.

### DIFF
--- a/arduino-ide-extension/src/browser/arduino-ide-frontend-module.ts
+++ b/arduino-ide-extension/src/browser/arduino-ide-frontend-module.ts
@@ -145,6 +145,8 @@ import { OutputToolbarContribution } from './theia/output/output-toolbar-contrib
 import { AddZipLibrary } from './contributions/add-zip-library';
 import { WorkspaceVariableContribution as TheiaWorkspaceVariableContribution } from '@theia/workspace/lib/browser/workspace-variable-contribution';
 import { WorkspaceVariableContribution } from './theia/workspace/workspace-variable-contribution';
+import { DebugConfigurationManager } from './theia/debug/debug-configuration-manager';
+import { DebugConfigurationManager as TheiaDebugConfigurationManager } from '@theia/debug/lib/browser/debug-configuration-manager';
 
 const ElementQueries = require('css-element-queries/src/ElementQueries');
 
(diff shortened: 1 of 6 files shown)


per1234 (commented):
I tested it out with my Zero, standalone CMSIS-DAP probe, and J-Link and it works fine, but the "No Configurations" experience is really not pleasant. It seems to me that we're trading the very minor annoyance of the `.theia` folder pollution of the sketch for a significantly worse user experience. My suggestion is to wait until a solution is found that can avoid the "No Configuration" thing from happening.

kittaakos:
Thank you for the review, Per.

> but the "No Configurations" experience is really not pleasant. It seems to me that we're trading the very minor annoyance of the `.theia` folder pollution of the sketch for a significantly worse user experience.

This is a bit more complex. Let assume we keep the `.theia` folder in the sketch folder, and we perform the following steps:
 - Create a new sketch. A sketch will be created in the temp folder. Let's say it's under `~/temp/x`.
 - Compile the new sketch. The compiled output will be under another temp folder: `~/temp/arduino-sketch-y`.
 - Start debug. (As discussed, we keep the `.theia` folder inside the sketch folder.) A file will be created under `~/temp/x/.theia/launch.json`, and the `executable` path inside the `launch.json` will point to `~/temp/arduino-sketch-y/board:fqbn/x.ino.elf`.
 - Save the sketch and pick a folder somewhere, let's say `~/Documents/Arduino/x`; here, it gets tricky. We could copy `~/temp/x/.theia/launch.json` to `~/Documents/Arduino/x/.theia/launch.json`, but the `executable` entry inside the `launch.json` still points to `~/temp/arduino-sketch-y/board:fqbn/x.ino.elf`. This can be invalid, and this is the issue here, not that we keep the `./theia` folder inside the sketch folder or not.

I hope this clarifies. If no, let me know.

kittaakos:
> My suggestion is to wait until a solution is found that can avoid the "No Configuration" thing from happening.

For the record; this is how it works now from the `main`, so technically, we do not change the UX at all.

ubidefeo (approved):
Thank you for explaining the case in which we would move the sketch and pre-existing config would have the wrong path.
I think how it is working now is good enough.
We'll investigate the debugger experience further down the road, especially since we'll have to find a more Arduino way of adding custom parameters via `debug_custom.json`.

This PR has therefore my approval.
Thank you @per1234 for your review 👍🏼

PR #190, opened 2021-03-09
[Toolbar] Hover style and Toggle State [ATL-1107][ATL-1045]

### LINKED ISSUE #173: Verify and Upload buttons do not change colour when activated.
Verify Button does not change colour when activated, yet.

fstasi:
## Why
- We are missing hover and pressed state on the buttons, which can be confusing to users switching from Java IDE.
- The Java IDE has the buttons remain “activated” until an operation is carried out.
- We want the "Verify" and "Upload" buttons (and menu items) to be disabled while the corresponding operation is in progress.

## How
- added "hover" behavior to all toolbar items, via CSS rule
- added "toggle" state to arduino toolbar buttons, by adding a prop, binded to the relative registry command. The `isToggled` function of the registry command was previously not called by Toolbar Items. I adopted the same mechanism used for `isEnabled`.
- set isToggled/isEnabled using an internal state updated by the verify and upload functions

## Other
[Jira Task (cosmetics)](https://arduino.atlassian.net/browse/ATL-1107)
[Jira Task (disable while in progress)](https://arduino.atlassian.net/browse/ATL-1045)
Fixes #173

### DIFF
--- a/arduino-ide-extension/src/browser/contributions/verify-sketch.ts
+++ b/arduino-ide-extension/src/browser/contributions/verify-sketch.ts
@@ -18,6 +18,8 @@ export class VerifySketch extends SketchContribution {
     @inject(BoardsServiceProvider)
     protected readonly boardsServiceClientImpl: BoardsServiceProvider;
 
+    verifyInProgress = false;
+
     registerCommands(registry: CommandRegistry): void {
         registry.registerCommand(VerifySketch.Commands.VERIFY_SKETCH, {
             execute: () => this.verifySketch()
@@ -27,6 +29,7 @@ export class VerifySketch extends SketchContribution {
         });
         registry.registerCommand(VerifySketch.Commands.VERIFY_SKETCH_TOOLBAR, {
             isVisible: widget => ArduinoToolbar.is(widget) && widget.side === 'left',
+            isToggled: () => this.verifyInProgress,
             execute: () => registry.executeCommand(VerifySketch.Commands.VERIFY_SKETCH.id)
         });
     }
@@ -65,7 +68,11 @@ export class VerifySketch extends SketchContribution {
     }
 
     async verifySketch(exportBinaries?: boolean): Promise<void> {
+
+        this.verifyInProgress = true;
         const sketch = await this.sketchServiceClient.currentSketch();
+        this.verifyInProgress = false;
+
         if (!sketch) {
             return;
         }
--- a/arduino-ide-extension/src/browser/contributions/verify-sketch.ts
+++ b/arduino-ide-extension/src/browser/contributions/verify-sketch.ts
@@ -22,13 +22,16 @@ export class VerifySketch extends SketchContribution {
 
     registerCommands(registry: CommandRegistry): void {
         registry.registerCommand(VerifySketch.Commands.VERIFY_SKETCH, {
-            execute: () => this.verifySketch()
+            execute: () => this.verifySketch(),
+            isEnabled: () => !this.verifyInProgress,
         });
         registry.registerCommand(VerifySketch.Commands.EXPORT_BINARIES, {
-            execute: () => this.verifySketch(true)
+            execute: () => this.verifySketch(true),
+            isEnabled: () => !this.verifyInProgress,
         });
         registry.registerCommand(VerifySketch.Commands.VERIFY_SKETCH_TOOLBAR, {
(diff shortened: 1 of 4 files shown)


arduino-ide-extension/src/browser/contributions/verify-sketch.ts
```diff
@@ -65,7 +71,17 @@ export class VerifySketch extends SketchContribution {
     }
 
     async verifySketch(exportBinaries?: boolean): Promise<void> {
+
+        // even with buttons disabled, better to double check if a verify is already in progress
```
kittaakos, line 75: > better to double check

Why? This code should not run if `isEnabled` is implemented correctly for the command handler. Did you experience a bug?

arduino-ide-extension/src/browser/contributions/verify-sketch.ts
```diff
@@ -18,15 +18,21 @@ export class VerifySketch extends SketchContribution {
     @inject(BoardsServiceProvider)
     protected readonly boardsServiceClientImpl: BoardsServiceProvider;
 
+    verifyInProgress = false;
```
kittaakos, line 21: This should be a `protected` field, not a `public` mutable property. Same for the upload part.

arduino-ide-extension/src/browser/contributions/verify-sketch.ts
```diff
@@ -18,15 +18,21 @@ export class VerifySketch extends SketchContribution {
...
     registerCommands(registry: CommandRegistry): void {
         registry.registerCommand(VerifySketch.Commands.VERIFY_SKETCH, {
-            execute: () => this.verifySketch()
+            execute: () => this.verifySketch(),
+            isEnabled: () => !this.verifyInProgress,
         });
         registry.registerCommand(VerifySketch.Commands.EXPORT_BINARIES, {
-            execute: () => this.verifySketch(true)
+            execute: () => this.verifySketch(true),
+            isEnabled: () => !this.verifyInProgress,
```
kittaakos, line 30: The Java IDE allows performing an `Export compiled binary` command. True, that the export does not start until the compile runs.

arduino-ide-extension/src/browser/style/main.css
```diff
@@ -15,15 +15,24 @@
     background: var(--theia-arduino-toolbar-background);
 }
 
-.p-TabBar-toolbar .item.arduino-tool-item > div:hover {
-    background: (--theia-arduino-toolbar-hoverBackground);
+.p-TabBar-toolbar .item.arduino-tool-item:hover > div {
+    background: white;
```
kittaakos, line 19: We cannot use hard-coded color so neither `white` nor `rgb(255,204,0)` is acceptable. We use VS Code themes, any new colors we want to use have to be derived from the available theme colors.
 - https://code.visualstudio.com/docs/getstarted/themes
 - https://code.visualstudio.com/api/extension-capabilities/theming
 - https://code.visualstudio.com/api/references/theme-color
 
Here is the Theia doc: https://github.com/eclipse-theia/theia/pull/6475/files#diff-06572a96a58dc510037d5efa622f9bec8519bc1beab13c9f251e97e657a9d4ed

kittaakos (commented):
It looks promising :+1: I added a few remarks. Also, please squash your commits; we need one single commit for the PR. Thank you!

arduino-ide-extension/src/browser/contributions/verify-sketch.ts
fstasi: I added this check in order to future-proof the verify-sketch implementation:

https://arduino.atlassian.net/browse/ATL-1045 reports an error (Error: Request upload failed with message: 2 UNKNOWN: exit status 1) when a verify/build is run while another is in progress.

At the moment we only have buttons/menu items, but the check whether or not continue with the action should be inside the action handler itself in my opinion.. 

Do you have concerns about this approach?

arduino-ide-extension/src/browser/contributions/verify-sketch.ts
fstasi: nice :)

fstasi:
> It looks promising 👍 I added a few remarks. Also, please squash your commits; we need one single commit for the PR. Thank you!

Is that ok to squash merge the PR in order to have a single commit on master? Or do you mean you want a single commit in the PR?

PR #203, opened 2021-03-11
ATL-974: Use `board search` command from the CLI

kittaakos:
What it does:
 - This PR removes any fuzzy-matching logic from the IDE and switches to the new `board search` command from the CLI.
 - From now on, the CLI dictates what's the result in the board select dialog.
 - Removes a couple of obsolete code/modules.

Please review. Thank you!


ubidefeo (approved):
Tested the IDE+CLI PR combo and it does what I expect
Thanks

PR #212, opened 2021-03-12
Added overwrite confirmation to ZIP lib install

kittaakos:
This PR adds support for overwrite confirmation when installing a lib from a zip.

Steps to verify:
 - Install a lib from a zip that was not installed,
 - Install it again,
 - You should see the confirm dialog whether to overwrite or not. If you hit `No`, nothing happens. If you hit `Yes`, the existing lib will be overwritten.

Please verify. Thank you!


arduino-ide-extension/src/browser/contributions/add-zip-library.ts
```diff
@@ -49,21 +50,60 @@ export class AddZipLibrary extends SketchContribution {
...
+            this.outputChannelManager.getChannel('Arduino').clear();
+            dialog.open();
+            await this.libraryService.installZip({ zipUri, overwrite });
+        } catch (error) {
+            if (error instanceof Error) {
+                const match = error.message.match(/library (.*?) already installed/);
+                if (match && match.length >= 2) {
+                    const name = match[1].trim();
+                    if (name) {
+                        throw new AlreadyInstalledError(`A library named ${name} already exists. Do you want to overwrite it?`, name);
```
per1234, line 83:
```suggestion
                        throw new AlreadyInstalledError(`A library folder named ${name} already exists. Do you want to overwrite it?`, name);
```
The library name is defined by the `name` field of `library.properties`, but this is about a folder name collision (two libraries that use the same installation folder name could have completely different `name` values).

I realize that this is following the convention established by the Arduino CLI error message, but since the message text is defined here in the IDE code base, I think it's worth fixing it here from the start.

per1234 (changes requested):
Really a great improvement.

I made a suggestion for modifying the confirmation dialog text, but the functionality is working perfectly for me.

kittaakos:
I have updated the PR and force-pushed the requested `library` -> `library folder` change. Please verify:

Screen Shot 2021-03-15 at 09 32 48

Thank you!

silvanocerza (approved):
Works great! :+1:

per1234 (approved):
Thanks Akos!

kittaakos:
Thank you for the reviews 👍

Issue #185, opened 2021-03-08
Sketchesd in alphabetical order but capitalised are grouped first


ubidefeo:
**Describe the bug**
When the sketchbook menu is expanded all the ones which have capitalised names come as a group in alphabetical order, followed by a group of examples with lowercase names as another group (see screenshot)

**To Reproduce**
1. **File > Sketchbook**
2. Scroll examples
3. Notice that uppercase named sketches are grouped first

**Expected behavior**
Sketches should be in true alphabetical order, despite their initial

**Screenshots**
<img width="295" alt="Screenshot 2021-03-08 at 16 13 25" src="https://user-images.githubusercontent.com/1894487/110340731-db86ef80-8029-11eb-9262-44948d527370.png">

**Desktop (please complete the following information):**
 - OS: Mac OS 10.15.7
 - Version: 2.0.0-beta.3 Date: 2021-02-26T10:47:47.411Z

silvanocerza:
I think that list is generated directly by the IDE, it's not a concern of the CLI as of now.

ubidefeo:
thanks @silvanocerza 
@kittaakos can you look into this?

Sebastian-Brzuzek:
I'm new here, so I'm not sure if proposed solution is reasonable, but if looks like assing "order" property with lower case "label" value in sketches/examples solves the issue without impacting other elements. (I have tested solution on Windows only)

ubidefeo:
Closed by #222

PR #213, opened 2021-03-12
ATL-1119 Disable code signing when workflows run from forks

rsora:
This PR supersedes https://github.com/arduino/arduino-ide/pull/208 and comes from a fork of this repo in order to test the workflow update it contains.

The purpose of this PR is to refactor the CI workflows to be more "fork friendly".
If workflows are run on a PR generated by a fork, where secrets are not available, the CI will run skipping the signing steps, providing unsigned artifacts.

What this PR does:

- Skip Mac/Win code signing and Apple notarization only if PR comes from a fork
- Disable workflows entirely if the user enabled Github Actions in
their fork repo
- Add steps to help Mac users to test their forked code in BUILDING.md

PR #222, opened 2021-03-15
Fixed sorting of sketches and examples - ignore case

Sebastian-Brzuzek:
simple solution for issue #185

### DIFF
--- a/arduino-ide-extension/src/browser/contributions/examples.ts
+++ b/arduino-ide-extension/src/browser/contributions/examples.ts
@@ -100,7 +100,7 @@ export abstract class Examples extends SketchContribution {
                     }
                 };
                 pushToDispose.push(this.commandRegistry.registerCommand(command, handler));
-                this.menuRegistry.registerMenuAction(submenuPath, { commandId, label: sketch.name });
+                this.menuRegistry.registerMenuAction(submenuPath, { commandId, label: sketch.name, order: sketch.name.toLocaleLowerCase() });
                 pushToDispose.push(Disposable.create(() => this.menuRegistry.unregisterMenuAction(command)));
             }
         }


kittaakos:
Thanks for your contribution, @Sebastian-Brzuzek. Your PR currently depends on #213. Please bear with us.
The fix seems good to me. Can we do the same for libraries? We have to change the code here:
https://github.com/arduino/arduino-ide/blob/5c8669d699b5f02251d04f338ea9b87999786893/arduino-ide-extension/src/browser/contributions/include-library.ts#L117

Sebastian-Brzuzek:
@kittaakos thanks for feedback. It looks like there is general problem with Action Menu in Theia. Take a look at this commit: https://github.com/Sebastian-Brzuzek/theia/commit/c06060722e957560bc9f4ed992fc98e1a20d2b57 

I can see the following options:

1. Convince Theia team to make global change - which will solve this kind of problems once for all.
2. Find all references where registerMenuAction method is used in Arduino Ide and to add precalculated `order` property as needed.
3. Add some kind of "wrapper" to `registerMenuAction` method (or `MenuAction`) to override current Theia implementation in one place for Arduino Ide project.

I can help with option 2 or 3. But I'm not sure if I will have time to make some progress with option 1.

One more question - what is expected order in case of menu items with labels which differ only on lower/upper case letter level? I know it is a little bit strange edge case 😉 

Maybe there are some other options, what do you think?

kittaakos:
For the sake of simplicity, I am going to approve this PR once you have rebased it from the `main`. Could you please do that? 😊 After the rebase, the workflow should restart and pass. #213 was merged.

Thank you so much for looking into the possible solutions.

> 1. Convince Theia team to make global change - which will solve this kind of problems once for all.

Let's do this. I can help with this.

> 2\. Find all references where registerMenuAction method is used in Arduino Ide and to add precalculated `order` property as needed.

We could do that, but once we have a Theia-based solution, it's unnecessary.

> 3\. Add some kind of "wrapper" to `registerMenuAction` method (or `MenuAction`) to override current Theia implementation in one place for Arduino Ide project.

Ideally, it should be possible to customize the default `MenuModelRegistry` in Theia. I will look into this as part of the first option.

Sebastian-Brzuzek:
It has taken some time (I have done rabase using the GitHub Desktop App for the first time :smiley: ) but finally checks are successful ;-)

Regarding customization of the default `MenuModelRegistry` in Theia, it is great idea for flexible solution, which will allow for safe transition without breaking changes.

kittaakos:
> It has taken some time (I have done rabase using the GitHub Desktop App for the first time 😃 ) but finally checks are successful.

I am glad you've managed to do it 💪 

I have checked your code, it looks good to me. I also tried your changes in-action, it's also good.

Screen Shot 2021-03-17 at 09 29 13

We need approval from someone else. @ubidefeo could you please verify the change and approve it if you like it? Thank you!

kittaakos (approved):
Thank you for your contribution, @Sebastian-Brzuzek 👍 Your changes look good to me.

ubidefeo (approved):
thank you for your contribution @Sebastian-Brzuzek 
sorry it took me a while to chime in with my 👍🏼  

🙏🏼

PR #243, opened 2021-03-19
Removed the min window height and width.

### LINKED ISSUE #216: Unable to shrink IDE window.
OS: Ubuntu 20.04
Display: 1366x768
IDE: 2.0.0-beta.3

Opening state of IDE is taller than my screen and will not shrink using normal methods. Because the IDE is too tall I can not see the bottom of the program.

Different topic but a maximize button would be nice.

kittaakos:
Made the boards config dialog slightly smaller.

Closes #216.


per1234 (changes requested):
This is really a great improvement.

I did encounter a problem though: If you have the package index URL for the popular 3rd party "DrAzzy" boards platforms in your preferences, but no installation of the ATTinyCore platform it provides, the "Ports" menu disappears from the "Select Other Board & Port" dialog:
Clipboard01

Notice that DrAzzy has made some creative usage of the package index's `packages[].platforms[].boards` field:
```json
"boards": [
  {
    "name": "<b>Program via ISP or Serial:</b> ATtiny841/441, ATtiny85/45/25, ATtiny84/44/24, ATtiny1634, ATtiny861/461/261, ATtiny167/87, ATtiny43, ATtiny828, ATtiny88/48, ATtiny4313/2313"
  },
  {
    "name": "<br/><b>USB (Micronucleus) Support:</b> DigiSpark (t85), Digispark Pro (t167), MH-ET (t88), Wattuino/Nanite/etc (t841), CaliforniaSTEAM (t84)"
  },
  {
    "name": "<br/><b>Windows users:</b> If USB drivers are not already installed, run the post_install.bat manually or DL from <a href=https://azduino.com/bin/micronucleus>https://azduino.com/bin/micronucleus</a>"
  }
],
```

### Steps to reproduce

1. (In Arduino IDE) Select **File > Preferences** from the menus.
2. Add the following URL to the "Additional Boards Manager URLs" field: `http://drazzy.com/package_drazzy.com_index.json` (separate with comma if there are multiple URLs in the field).
3. Click the "OK" button.
4. Close the Arduino IDE.
5. Start the Arduino IDE
6. Click the drop down board selection menu in the button tool bar.
7. From the menu, click on "Select Other Board & Port" - There is no "Ports" menu in the dialog that opens.

Note that you must not have the ATTinyCore boards platform installed. If you have it installed, do an uninstallation before restarting the IDE. The reason is that the IDE only uses the strings from the `packages[].platforms[].boards` field when the platform is not installed. After installation, it uses the board name strings from boards.txt instead, which are different in the case of ATTinyCore.

This issue does not occur with Arduino IDE 2.0.0-beta.4

I was able to reproduce this on both Windows and Linux machines.

ubidefeo:
@per1234 
since it does not occur on Beta 4, shouldn't we declare this solved?

per1234:
@ubidefeo no. The fact that the issue I reported does not occur in 2.0.0-beta.4 proves that it was introduced by this PR.

ubidefeo:
@per1234 
Oh, right!
my synapses got crossed there.
thanks 👍🏼

kittaakos:
> I reported does not occur in 2.0.0-beta.4

I can reproduce it with `beta.4`. Although the ports are still visible, the layout is broken. Anyways, I will fix it as part of this PR.

screencast 2021-03-29 13-14-25

kittaakos:
I have updated the dialog stylings. I hope it provides a somewhat better UX. Could you please take another look, Per? Thank you!

https://user-images.githubusercontent.com/1405703/112850059-f4734580-90a9-11eb-916b-a6acc24769cd.mp4

per1234 (approved):
Works great for me now. I was personally affected by the previous minimum window dimensions because my supplemental laptop's display is smaller than what was necessary to provide for easy usage of the IDE.

Thanks Akos!

PR #264, opened 2021-03-24
ATL-1106: Made all non-workspace editors read-only.

kittaakos:
What it does:

This PR makes all editors that are not in the current workspace (sketch folder) read-only.

Try the `Peek` > `Peek Definiton` functionality on `Serial#begin` and `bar`. (Note the cursor location: `<|>`):

`Sketch.ino`:
```cpp
#include "other.h"
void setup() {
  ba<|>r();
  Serial.begi<|>n(9600);
}
void loop() { }
```

`Other.h`:
```cpp
void bar() {}
```

`Serial#begin` should open in a read-only embedded editor. You should be able to edit in the embedded editor if you peek at the definition of `bar`. Opening an editor should behave the same as the embedded peek editor. Please review. Thank you!


per1234 (approved):
This is really great Akos. Thanks!

I did find a related problem, which is that the tab control menu allows me to delete or rename the files opened via "Go to Definition". I could see a user using "delete" to remove what appears to be an unwanted file from their sketch. But I'm guessing that is out of scope for this PR, so I'm giving a resounding thumbs up on this.

ubidefeo:
Downloading now, but I share @per1234 's opinion that non-workspace files should not be deleted.
I see one problem ahead: are files in the Sketch's `src` folder considered Workspace file?
Those ones should be editable and deletable.
Can of worms, anyone?

kittaakos:
> opinion that non-workspace files should not be deleted.

It works like that.

> are files in the Sketch's `src` folder considered Workspace file?

Yes. They're part of the Theia/VS Code workspace (== current sketch folder.)

> Those ones should be editable and deletable.

You should be able to delete and modify them. If no, it's a bug.

kittaakos:
> > opinion that non-workspace files should not be deleted.
> 
> I open a follow-up.

Here is the follow-up: https://github.com/arduino/arduino-ide/issues/274

ubidefeo (approved):
Tested and working

PR #267, opened 2021-03-24
ATL-1150: Bumped the `cortex-debug` to `0.3.10`.

kittaakos:


kittaakos:
CCing @fstasi to share how to update a VS Code extension in the new IDE.

fstasi (approved):
Thanks! Seems straightforward...I can write some lines in a doc about this.

ubidefeo (approved):
I don't have ways to test this one for current lack of hardware.
Also it would require a custom configuration.
The change is clearly gonna bring in a new version of this extension which does not seem to have bug-fixes over `0.3.10`.
Be aware that soon we'll have to update to `0.3.12` since now they're in pre-release.
This adds 
>A new runToEntryPoint configuration property to replace runToMain allowing the actual entry point name to be configured

which will allow us to set `setup()` as our point of entry for debug 👍🏼

PR #276, opened 2021-03-26
ATL-837: board serial number

fstasi:
Add serial number to Board Info

Task [ATL-837](https://arduino.atlassian.net/browse/ATL-837)


kittaakos:
VID/PID can be acquired from the `Port#properties`. See #1335.

kittaakos:
Currently, IDE2 can show the board number, VID, and PID, but the serial number is missing. Follow-up: https://github.com/arduino/arduino-ide/issues/1435

PR #285, opened 2021-03-29
chore: update CLI version, gRPC and bit of documentation

fstasi:
As per title,

- updating CLI to 29 march nightly
- updating gRPC interfaces
- add bits of documentation about the re-generation of the gRPC stubs


kittaakos (approved):
Thank you!

PR #294, opened 2021-03-31
ATL-1145: Suppress error if Git is not on `$PATH`.

kittaakos:
This PR does one thing: it suppresses the error message when the Git executable is not on the `$PATH`.

How to verify: this is an interesting part, I did the straightforward way and renamed all my Git executables to something else so that `git version` fails from a terminal. I renamed the followings:
 - `C:\Program Files\Git\cmd\git.exe`,
 - `C:\Users\my_name\scopes\shims\git.EXE`, and
 - made sure `/mingw64/bin/git` (for Git Bash) also not accessible.

Once you have a setup without git on the `$PATH`, start the app. You should not see any errors.


rsora (approved):
Tested in a clean Win environment:
- latest nightly: I can see the error message
- Build from this PR: I can't see the error anymore
👍

kittaakos:
Thank you for the review.

PR #305, opened 2021-04-09
ATL-786: Progress indication for install.

kittaakos:
This PR gets rid of the modal dialogs when installing/uninstalling libs and platforms.


silvanocerza:
Just tested and I like this enhancement, my only concern is that probably the bar might be too small, could it be made a bit thicker or the size is fixed?

kittaakos:
> ould it be made a bit thicker or the size is fixed?

Yes, we can do that. Also, the default color should not be the hardcoded blue from Theia but aligned with the status bar color. I think it is blue.

I wanted to reset the progress bar between the tasks (there is no progress for `task`s, only for `download`s), and use an _unknonwk_ progress indicator, but it does not seem to work. (Related: https://github.com/eclipse-theia/theia/discussions/9323)

kittaakos:
> Yes, we can do that.

> and use an _unknonwk_ progress indicator

I have updated the PR. Here it is in action. Now, the progress bar can jump between unknown and normal mode:

screencast 2021-04-09 16-56-21

silvanocerza (approved):
Love it, it's so much better now!

silvanocerza (approved):

kittaakos:
I am merging this into the other PR.

PR #342, opened 2021-04-27
Refactor Init process to use new CLI gRPC API

silvanocerza:
The CLI initialization process has been changed in significant ways with PR arduino/arduino-cli#1274, it needs to be thorougly tested before merging to avoid breaking the IDE.


arduino-ide-extension/src/node/core-service-impl.ts
```diff
@@ -26,6 +26,7 @@ export class CoreServiceImpl extends CoreClientAware implements CoreService {
         const { sketchUri, fqbn, compilerWarnings } = options;
         const sketchPath = FileUri.fsPath(sketchUri);
 
+        await this.coreClientProvider.initialized;
```
fstasi, line 29: @silvanocerza what happens if the `initialized` fails or rejects? is it ok to proceed calling `coreClient()`?

silvanocerza: I think it's fine, if the `Init` gRPC function fails it means it has been called, so `Create` was successful meaning we have an internal instance in the CLI daemon and we have a "working" client.

If `Init` fails completely though we must have some other huge problems, with the refactoring failure to load a platform or library just logs an error, but it doesn't stop completely.

I still don't like how I implemented this though, I'd like to show the user a notification when there are this kind of failures but am not sure how to handle that.

ubidefeo:
@silvanocerza 
Installing a platform rebuilds the tools>board menu, but uninstalling does not

- install a platform
- look at the boards menu: it's there
- uninstall the platform
- look at the boards menu: it's still there

image

silvanocerza:
@ubidefeo I investigated a bit, the same issue is also present when uninstalling a library.

It seems to me this is a race condition, the list of boards divided by platform is retrieved before the platform is uninstalled, the strange thing though is that after installing a new one I'd expect it to stop showing the uninstalled platform but it doesn't. :thinking: 

This requires some further investigation I guess.

silvanocerza:
Rebased to fix conflicts and trigger a new build after updating the version of the Arduino CLI.
The issues with core/platforms installation is solved, it was an issue in the CLI after all.

ubidefeo (approved):
been poking around for almost an hour and haven't found anything which isn't already an issue in the Beta 7.
LGTM

PR #356, opened 2021-05-06
New Linux AppImage Launcher for gnomel

### LINKED ISSUE #641: How should this arduino-ide be installed on linux? Missing install script.
I've downloaded and extracted it in /opt
There is no install script like in the older IDE.
Running ./arduino-ide, doesn't launch anything and doesn't print error messages.

ArduinoShop:
The following files would need to be added to the Linux package:

1). Linux_Install.sh
2). desktop.template
3). Linux_Install_ReadME.txt
4). arduino2.png  --  this is my icon that I used for testing - you may have much better ideas...

If you need these don in Spanish or French let me know...


joshuag1000:
Ideally an appdata xml file should also be provided.
like the one i have used here: https://github.com/joshuag1000/flathub/blob/cc.arduino.arduinoide2/cc.arduino.arduinoide2.appdata.xml

Linux_Install.sh
```diff
@@ -0,0 +1,227 @@
...
+		fi
+		
+		mkdir -p "${HOME}/.local/bin"
+		rm -f ${HOME}/.local/bin/${RESOURCE_NAME}
+		ln -s ${HOME}/.local/lib/${RESOURCE_NAME}/arduino-ide ${HOME}/.local/bin/${RESOURCE_NAME}
+	
+		# Create a temp dir accessible by all users
+		TMP_DIR=`mktemp --directory`
+
+		# Create *.desktop file using the existing template file
```
hfiguiere, line 121: the icon and desktop file should follow the proper app-id convention, like the one used for flathub.

See https://specifications.freedesktop.org/desktop-entry-spec/latest/ar01s02.html#desktop-file-id

The id should be `cc.arduino.arduinoide2` as suggested for the flathub submission.
Note that if you app register an app-id for DBus use (like Qt or Gtk do) then that's the one that should be used and coordinate with the flathub submission (https://github.com/flathub/flathub/pull/3176 ) to changed it there too.

With that the .desktop should be `cc.arduino.arduinoide2.desktop` and the icon should be `cc.arduino.arduinoide2.png`. Note: if you have a SVG icon it would be so much better.

A 48x48 icon is way too small. 128x128 is the current standard.

Linux_Install.sh
```diff
@@ -0,0 +1,227 @@
...
+		sed -e "s,<BINARY_LOCATION>,${EXE_PATH}/${RESOURCE_NAME},g" \
+	    -e "s,<id>,local,g" \
+        -e "s,<ICON_NAME>,${LIB_PATH}/arduino2.png,g" "${SCRIPT_PATH}/desktop.template" > "${TMP_DIR}/${RESOURCE_NAME}l.desktop"
```
hfiguiere, line 124: instead of using `sed` you should use `destkop-file-edit`. It's way easier to understand and exactly designed for that. There is also `desktop-file-install` to install it.

hfiguiere (commented):
I come for the XDG and flatpak perspective, since this is to some extent closely related, getting it right from start is probably the best.

Linux_Install.sh
joshuag1000: Just a note the flatpak pr request was updated to change the id to `cc.arduino.IDE2` to match what is used here: https://github.com/arduino/arduino-ide/blob/0a87fd00f38c923a36ce22d9a787fd071a6173e7/electron/packager/config.js#L137

hfiguiere: then that's what shall be used here too.

ArduinoShop:
Hi - I have been working with the Arduino team to write "Linux Launcher" scripts. The idea is to be able to Launch IDE 2.0 from icons in a Gnome dock..

I have the scripts posted on my github...

https://github.com/ArduinoShop/arduino-ide/tree/main/AppImage%20Launcher

I am not an adept at using git so I am not sure how to "merge" them with Arduino's efforts.

ben-sembira-1:
Hey everyone,

Is this PR still active? If not, I’d be happy to pick it up and continue the work or start fresh if needed. 😄

I think it’s a fantastic idea to include desktop files and installation scripts for the Arduino IDE in future distributions. I've actually developed a similar solution for my workspace and wanted to share it with the community. That’s when I stumbled upon this PR by accident.

Here are the scripts I’ve created: https://github.com/ben-sembira-1/workspace/tree/master/scripts/arduino.

I’m not sure if this should be part of the same PR, but I believe aiming for a stable APT release could significantly enhance installation stability and accessibility on Linux systems. These two efforts seem closely related as they both target easier and more stable installations on Linux.

I also found issue #2138 about creating an APT package for the Arduino IDE, but it seems to lack recent activity. While there is an APT package for Arduino, it appears to only support IDE 1.x versions. I couldn’t find an equivalent for IDE 2.x.

Looking forward to your thoughts!

PR #392, opened 2021-05-27
Footer min-height for library and board manager

IrregularShed:
On Windows 10 (and possibly others), rolling over the items in the Library Manager and Boards Manager causes an unpleasant jump, as seen here:

ezgif com-gif-maker

This PR increases the `min-height` of the footer for the list items from 26px to 30px, which is the minimum value required to prevent this from occurring, and improves the UX in doing so, without changing the spacing excessively:

ezgif com-gif-maker (1)


kittaakos:
I have verified the changes locally on macOS. It works as described. Thank you for the patch 👍 

+1 for merging it.

IrregularShed:
Is there any chance of this being merged? I've just installed beta 10 and this issue hasn't been fixed by any other changes. It's quite jarring - I'm autistic and the jump feels painful (which was my motivation for finding a fix in the first place!)

PR #415, opened 2021-06-16
force electron-builder version resolution

fstasi:
## Why
arduino-ide uses on node v12. Latest electron-builder release requires node v14. This makes electron-builder unable to build the ide.

## How
Forcing electron-builder version and resolution in the package.json


silvanocerza (approved):

rsora:
:+1:

PR #423, opened 2021-06-22
eslint to fix unused imports

fstasi:
### Why
tsc complains about unused imports, stopping the builds in the CI. We want to catch unused imports asap.
Eslint supports unused-vars rule, but this appears to be too strict for our usecase. We need to catch imports *only*, and possibly auto-fix those before committing.

### How
[eslint-plugin-unused-imports](https://www.npmjs.com/package/eslint-plugin-unused-imports) integrates with eslint, detecting unused imports and removing them from the codebase. This automaically integrates with husky/lint staged that are already in place

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -1,12 +1,4 @@
 module.exports = {
-    ignorePatterns: [
-        'node_modules/*',
-        '/*',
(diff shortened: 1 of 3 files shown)


AlbyIanna (approved):
LGTM

PR #450, opened 2021-07-12
ATL-1451: reveal sketch directory in file explorer

fstasi:
## Why
Reveal sketchbook directory in explorer/finder is not working. This is caused by the `node.id` not being a valid URI.

## How
Correctly parse the `node.uri` to a valid, independent, path, via `fileService.fsPath`. Then use `remote.shell.openPath(fsPath);` to open the sketch directory

## Notes
Jira Task:
- [[sketchbook] Open in folder on windows does not work](https://arduino.atlassian.net/browse/ATL-1451)


AlbyIanna (approved):
LGTM

PR #459, opened 2021-07-20
Improve remote sketchbook tree

fstasi:
## Why
A major re-design of the sketchbook tree is needed in order to enable a number of features.

## What to expect
- improved performances when refreshing the sketch tree. Resources are now loaded *on demand* meaning you will notice a short delay when opening a directory for the first time. A caching system is in place in order to make subsequent actions more quick
- The sketchbook tree highlights the file currently marked as "active" in the editor. The containing folders are automatically expanded if closed, in order to help the users immediately see the location of the files they are working on
- Creating a file in the local file system automatically adds it to the sketchbook tree and make it selectable. This is required in order to allow creation/removal/rename of files in the "local copy" of a remote sketchbook
- overall performance improvements: mapping between remote sketches and local files is done once, as soon as the cache is generated. This ensures less access to the fileSystem and hence a snappier experience of the feature

### DIFF
--- a/arduino-ide-extension/src/browser/create/create-api.ts
+++ b/arduino-ide-extension/src/browser/create/create-api.ts
@@ -4,6 +4,7 @@ import { posix } from './create-paths';
 import { AuthenticationClientService } from '../auth/authentication-client-service';
 import { ArduinoPreferences } from '../arduino-preferences';
 import { SketchCache } from '../widgets/cloud-sketchbook/cloud-sketch-cache';
+import { Create, CreateError } from './typings';
 
 export interface ResponseResultProvider {
   (response: Response): Promise<any>;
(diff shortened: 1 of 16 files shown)


arduino-ide-extension/src/browser/create/create-api.ts
```diff
@@ -32,48 +34,24 @@ export class CreateApi {
...
-  }): Promise<Create.Sketch | undefined> {
-    let url;
-    if (opt.id) {
-      url = new URL(`${this.domain()}/sketches/byID/${opt.id}`);
-    } else if (opt.path) {
-      url = new URL(`${this.domain()}/sketches/byPath${opt.path}`);
-    } else {
-      return;
-    }
+  async sketch(id: string): Promise<Create.Sketch | undefined> {
```
AlbyIanna, line 53: is there a reason why this could possibly return undefined?

AlbyIanna (approved):
LGTM

PR #466, opened 2021-07-27
[ATL-1539] Integrate FWUploader into IDE2

fstasi:
### Why
We need to bundle the firmware uploader into the IDE2

### How
- added download script
- updated `package.json` accordingly
- exposed the fw uploader executable  via `executable-service-impl.ts`


AlbyIanna (approved):
LGTM

PR #469, opened 2021-08-05
[ATL-1533] Firmware&Certificate Uploader

fstasi:
### Why
Enable the users to update firmware and add certificates to boards

### How
- bundled the arduino firmware uploader binary into the ide2 package
- adding firmware uploader modal
- adding certificate uploader modal
- updated modals look&feel


ubidefeo (approved):
LGTM

PR #493, opened 2021-08-31
[ATL-1436] Re-center refresh icon when spinning

AlbyIanna:
Since we are migrating all icons to FontAwesome, I did it with the refresh icon and this solved the issue
EZ


rsora (approved):
Nice!

PR #494, opened 2021-08-31
[ATL-1571] Fix editor quick suggestions preference

AlbyIanna:
We were getting `quickSuggestion` instead of `quickSuggestions` :')


AlbyIanna:
@fstasi I agree, I'll put a little more effort into this and try to solve this

PR #497, opened 2021-09-01
[ATL-1294] At first ide startup invoke installation of arduino:avr

AlbyIanna:
## HOW
Use `LocalStorageService` to know if it's the first IDE startup, and in that case launch `arduino:avr` installation.

### DIFF
--- a/arduino-ide-extension/src/browser/arduino-frontend-contribution.tsx
+++ b/arduino-ide-extension/src/browser/arduino-frontend-contribution.tsx
@@ -11,6 +11,7 @@ import {
   ContextMenuRenderer,
   FrontendApplication,
   FrontendApplicationContribution,
+  LocalStorageService,
   OpenerService,
   StatusBar,
   StatusBarAlignment,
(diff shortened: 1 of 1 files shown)


rsora (approved):
Win :ok: 
Linux :ok:

ShineUnM:
Maybe, Let user to choose installing the AVR core lib or not is better?
For example, I never use AVR, now I have to wait it installing and remove it, and install my target core lib

fstasi:
> Maybe, Let user to choose installing the AVR core lib or not is better?
> For example, I never use AVR, now I have to wait it installing and remove it, and install my target core lib

@ShineUnM this PR is meant to close the gap between the IDE2 beta and the current IDE. The latter ships with AVR core pre-installed in order to make the experience as forgiving as possible for newcomers.

I'll discuss your comment with the UX team and evaluate if we can find a suitable solution for more advanced users.

Feel free to open an issue so that we can track your request and collect other feedback.

thanks a lot

PR #513, opened 2021-09-23
Implement unit tests for boards-auto-installer

AlbyIanna:
The first step for enabling us to implement unit tests on the frontend. Simply run `yarn test` 😄 

We decided to use [typemoq](https://github.com/florinn/typemoq) in order to mock our injectable services.

While doing the tests, we noticed the `ResponseService` were not being injected correctly as we were injecting `ResponseServiceImpl` directly and this was causing troubles, so we created a new interface `ResponseServiceArduino`.

### DIFF
--- a/arduino-ide-extension/src/browser/boards/boards-auto-installer.ts
+++ b/arduino-ide-extension/src/browser/boards/boards-auto-installer.ts
@@ -7,10 +7,9 @@ import {
   Board,
 } from '../../common/protocol/boards-service';
 import { BoardsServiceProvider } from './boards-service-provider';
-import { BoardsListWidgetFrontendContribution } from './boards-widget-frontend-contribution';
 import { BoardsConfig } from './boards-config';
-import { Installable } from '../../common/protocol';
-import { ResponseServiceImpl } from '../response-service-impl';
+import { Installable, ResponseServiceArduino } from '../../common/protocol';
+import { BoardsListWidgetFrontendContribution } from './boards-widget-frontend-contribution';
 
 /**
  * Listens on `BoardsConfig.Config` changes, if a board is selected which does not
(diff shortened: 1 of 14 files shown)


arduino-ide-extension/src/browser/boards/boards-auto-installer.ts
```diff
@@ -106,7 +106,7 @@ export class BoardsAutoInstaller implements FrontendApplicationContribution {
                 });
                 return;
               }
-              if (answer) {
+              if (answer === 'Install Manually') {
```
fstasi, line 109: this change is suspicious and not test related. Did we mess up with commits?

package.json
```diff
@@ -48,6 +53,16 @@
...
+  "mocha": {
+    "require": [
+      "reflect-metadata/Reflect",
+      "ignore-styles"
+    ],
+    "reporter": "spec",
+    "colors": true,
+    "watch-extensions": "js",
+    "timeout": 10000
+  },
```
fstasi, line 65: this configuration is kind of duplicated, I think we should stick with just one config (+ the launch.json) and find a way to attach FE tests to the yarn test command that gets called by the CI.

fstasi (changes requested):

arduino-ide-extension/src/browser/boards/boards-auto-installer.ts
AlbyIanna: No, I changed this on purpose. In the current implementation, we are checking if the answer is `'Yes'`: if so, we install the library; in any other case, we open the library manager. I didn't like that, because if we'd add another answer like `'Nah, I don't want to'`, we don't want to open it.

fstasi (approved):
💯 great PR

PR #524, opened 2021-10-05
Improve Serial Monitor Performances

### LINKED ISSUE #79: Serial Monitor update rate issue.
The Serial Monitor built in to the IDE has a significantly delayed update rate.

I am using the Arduino Zero. I am using the following to print serial information to the internal Serial Monitor and an external bit of software (GTKTerm v1.1) to monitor the serial information being written to the native USB port. 

` Serial.begin(115200);
  SerialUSB.begin(115200);`

1. Open built in Serial Monitor.
2. Open external Serial Monitor.
3. Open scope to view which output is currently active.
4. View lag in the built in Serial Monitor.
5. Compare the lag with an external Serial Monitor.

I would expect to see the update rate of the built in Serial Monitor match the modulated outputs of the microcontroller. I am using a scope to view the outputs of the Zero. I can see in real time which outputs are on compared to the built in Serial Monitor and the external Serial Monitoring software.

(Please see attached screenshots for context.)
no_cyan_pulses
no_yellow_pulses
no_pink_pulses

**Version Information:**
`Linux 5.8.0-31-generic arduino/arduino-pro-ide#33-Ubuntu SMP Mon Nov 23 18:44:54 UTC 2020 x86_64 x86_64 x86_64 GNU/Linux` 
`arduino-pro-ide_nightly-20201211_Linux_64bit`

**Additional Information Regarding Issue**
When I hold the board in reset, the built in Serial Monitor continues to update as if still reading incoming serial data.

-cl330b

### LINKED ISSUE #391: Serial monitor prints data with the wrong timestamp
**Describe the bug**

The Serial Monitor doesn't print the value just received with the correct timestamp but with the one of the previous received value.

If the board sends `A` at time `1` and `B` at time `2` this would be the output, first:

```
1 -> A
1 ->
```

then:

```
1 -> A
1 -> B
2 ->
```

Changing the baud rate to a lower or higher one doesn't change a thing.

**To Reproduce**
1. Create a Sketch with the following code:
```
void setup()
{
  Serial.begin(2400);
  while (!Serial);
}

void loop()
{
  static byte x = 0;
  Serial.println(x);
  delay(10000);
  x++;
}
```
2. Upload it to your board
3. Open the Serial Monitor
4. Set the baud rate to 2400 
5. Click Toggle Timestamp if not already on

Notice how the timestamp for the value printed now is not correct but reference to the previous one.
Also the first value is printed two times.

**Expected behavior**

I expect the values to be printed on the correct timestamp when it's received by the PC.

**Screenshots**

https://user-images.githubusercontent.com/3314350/119687560-9b285800-be47-11eb-9356-8eb83b02d07f.mp4

**Desktop (please complete the following information):**
 - OS: Linux
 - Version: 2.0.0-beta.7

**Additional context**

This issue has been original reported on the Arduino Forum: https://forum.arduino.cc/t/serial-monitor-gives-wrong-time-stamp

### LINKED ISSUE #519: Serial Monitor lags behind and uses a lot of CPU
When printing data to Serial at high speed (115200) the Serial Monitor cannot keep up with the board output and CPU performances jump through the roof.

OS: Mac OS 11.3

fstasi:
## Why
Serial monitor is affected by performance issues: basically it's not able to keep up with the grpc messages arriving from the CLI when a decent amount of message (>100/sec) is being sent from the board.

There are a number of reasons causing this undesirable effect:
- the communication between Frontend and Backend is very chatty: the FE basically pulls a message at a time from the BE and gets notified when a new message arrives (see [comment](https://github.com/arduino/arduino-ide/issues/519#issuecomment-934164852))
- The frontend re-renders the whole list of messages every time a new one is pulled
- JRPC is used to marshal/unmarshal the data, causing additional overhead to the messages
- all messages are aggregated (and then printed), without any limit to the maximum number

## How
A new architecture for the communication has been designed:
- the backend collects the messages from the CLI and aggregates them in array.
- every 32 milliseconds the BE sends a chunk of message to the FE, containing all messages arrived in the 32msc timeframe. The reason behind the "32" milliseconds is the refresh rate of the electron application (32msec ~= 30fps). 16ms would guarantee a 60fps, but at the cost of more CPU and reduced responsiveness in the IDE2 GUI
- the communication between FE and BE is now a "raw" WebSocket connection, without the overhead of JRPC.
- as the frontend receives the chunk of messages, it adds them to its internal state, assigning them a timestamp and removing the old ones when the buffer size is reached
- the frontend renders only the messages that are actually in the monitor window: this allows to dramatically reduce the CPU usage as only a few messages are printed, rather than millions

## How to test
- upload a sketch with `Serial.print` or `Serial.println` and use no `delay` in order to have a high throughput of data flowing from the board
- open the serial monitor in the IDE2

## Related issues
fixes #391, fixes  #79, fixes #519


AlbyIanna (approved):
🚀

PR #528, opened 2021-10-06
Theia 18 hotfixes

fstasi:
## What
- monaco suggestions are now correctly highlighted
- re-enabled editor tabs rename/delete
- avoid duplication of monaco tabs removing legacy ones (without the counter flag)


AlbyIanna (approved):
🌶️

PR #550, opened 2021-10-14
Add dialog to insert user fields for board that require them to upload

silvanocerza:
This PR adds a dialog for the user to enter custom fields necessary to upload to certain boards that require them.

If the user clicks the upload button on the toolbar and never upload to that board on that addres in the current IDE session the dialog will be shown, if all the fields are filled and the user clicks the Upload button the values will be cached for the current session. Subsequent uploads triggered with the upload button won't ask the user again to enter the fields but will use the cached ones.
If the user wants to change the fields after the first upload she must open the dialog from `Sketch > Configure And Upload` menu, edit the fields and click the Upload button again.

In any case clicking the Cancel button doesn't save the entered values.

User fields are not saved when closing the IDE.

Example:

https://user-images.githubusercontent.com/3314350/137513861-14b0b260-1b4d-41b6-ba85-37109596c16c.mp4


ubidefeo (approved):
Functionality tested and confirmed working

joyel24:
@silvanocerza Hello, thanks for contributing to the project but please, can you please fix #1488 as you implemented this features that causes troubles to everyone trying to upload sketch via OTA on their ESP32 devices without password.
Or even add the ability to enter empty field and confirm.

PR #611, opened 2021-11-16
Disable Breadcrumbs

fstasi:
## Why
In the last update of Theia we included breadcrumbs just below the tab name. This is an optional feature that we do not plan to support, hence needs to disabled by default. Advanced users can still enable breadcrumbs using the **unsupported** theia preferences via theia preference panel.

## How
Set the breadcrumbs to be disabled by default

### DIFF
--- a/browser-app/package.json
+++ b/browser-app/package.json
@@ -44,7 +44,8 @@
             "other": false,
             "comments": false,
             "strings": false
-          }
+          },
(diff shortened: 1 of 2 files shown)


ubidefeo (approved):
LGTM
thanks

ubidefeo:
PR approved

PR #612, opened 2021-11-16
Fix tabs ordering

fstasi:
## Why
After the latest Theia inclusion, tabs order gets scrambled. This is particularly true when the Sketch has multiple .ino files and/or a file different from the main .ino file was selected the last time the sketch was opened

## How
- Changed the cleanup script that finds duplicated tabs in monaco
- removed the now useless force open on the main .ino file. This was actually causing the issue as forcing the open of the file that was already opened caused the close of the original one and a new open in the last position


ubidefeo (approved):
The correct behaviour has been restored.
LGTM

ubidefeo:
I have reviewed/approved the PR
cc @per1234

PR #614, opened 2021-11-17
Rework listing of discovered ports

silvanocerza:
This PR does several things:
* Minimal refactoring and removal of old code
* Change ports ordering in Tools > Ports menu
* Change ordering of board and port combination in Board Selector dropdown list
* Change ordering of ports in Select Board dialog
* Change style of Board Selector dropdown list

---

Ports in Tools > Ports menu are now sorted like this:
1. Serial with recognized boards
2. Serial with unrecognized boards
3. Network with recognized boards
4. Network with unrecognized boards
5. Other protocols with recognized boards
6. Other protocols with unrecognized boards

All recognized boards name are shown in parentheses after the port name. 
In the screenshot below you see two boards for each port because I have development cores in my Sketchbook folder so two boards with different FQBN but identical name are found.

image

---

Board and ports combination in Board Selector dropdown are now sorted like this:
1. Serial with recognized boards
2. Serial with guessed boards
3. Serial with incomplete boards
4. Network with recognized boards
5. Other protocols with recognized boards

Previously selecting an item would change the ordering by putting it on top, now it doesn't anymore.
Network ports were never shown previously, only serial ports where.

image image

---

Ports in Select Board dialog are now sorted like this:
1. Serial with recognized boards
2. Serial with unrecognized boards
3. Network with recognized boards
4. Other protocols with recognized boards

If the Show all ports box is checked they're sorted like this:
1. Serial with recognized boards
2. Serial with unrecognized boards
3. Network with recognized boards
4. Network with unrecognized boards
5. Other protocols with recognized boards
6. Other protocols with unrecognized boards

image
image

---

Note: For testing I've used the refactored version of the [`mdns-discovery`](https://github.com/arduino/mdns-discovery/) that is still not released. So you'll have a different experience than me with network ports, you'll see fewer, they'll never disappear after being discovered and similar issues.


91volt (approved):

fstasi (approved):
Code is good (nice job!) but I need the updated index to properly test

PR #635, opened 2021-11-26
Remove language server log preference

AlbyIanna:
## What
Remove the `Enable language server logging` option in the preferences

## Why
This option was needed in the early development and testing phase of the Arduino Language Server. However, this is not something the target user will ever have a need for, so it only makes the IDE less friendly.


per1234 (changes requested):
Please preserve support for enabling language server logging by setting the `arduino.language.log` settings key manually in `settings.json`.

The logging capabilities can still be useful for investigating reports of problems with the Arduino Language Server. In that case, we can guide the user in enabling the logging via the IDE's secondary advanced UI (or just using a text editor on `settings.json`). So there is no need to have a control for this via the IDE's primary UI, and this is why it should be removed from the **File > Preferences** interface.

AlbyIanna:
@per1234 thank you for the hint. I've just re-enabled the option, now it can be toggled in the advanced settings but it's not shown in the main **Preferences** diaog.

fstasi (approved):

PR #663, opened 2021-12-03
Automatically install 'Arduino_BuiltIn' library at first startup

AlbyIanna:


per1234 (changes requested):
The "Arduino_BuiltIn" library is being installed, but its dependencies are not.

The library installation must be configured to install dependencies because the "Arduino_BuiltIn" library is only a dummy library that serves as a container for the list of "built-in" libraries.

The working change should result in the installation of this list libraries on the first startup of the IDE:
https://github.com/arduino-libraries/Arduino_BuiltIn/blob/main/library.properties#L10

per1234 (approved):
Thanks @AlbyIanna!

PR #673, opened 2021-12-09
IDE to run CLI with auto assigned port

fstasi:
## Why
It's not possible to make assumptions on what ports are available on the different OS the IDE can be run, and the default port 50051 is already in use in some cases (such as OS Monterey).
This issue is particularly annoying for first-time users that should go into the CLI configuration and change the port manually, before running the IDE.

## How
The IDE now force the CLI daemon port to `0` (automatically assigned from the OS). This ensure an available port is always picked up and the IDE can start normally, without manual intervention of the user

### DIFF
--- a/arduino-ide-extension/src/test/node/arduino-daemon-impl.test.ts
+++ b/arduino-ide-extension/src/test/node/arduino-daemon-impl.test.ts
@@ -2,21 +2,17 @@ import * as fs from 'fs';
 // import * as net from 'net';
 import * as path from 'path';
 import * as temp from 'temp';
-import { fail } from 'assert';
 import { expect } from 'chai';
 import { ChildProcess } from 'child_process';
 import { safeLoad, safeDump } from 'js-yaml';
-import { DaemonError, ArduinoDaemonImpl } from '../../node/arduino-daemon-impl';
+import { ArduinoDaemonImpl } from '../../node/arduino-daemon-impl';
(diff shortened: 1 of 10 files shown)


arduino-ide-extension/src/test/node/arduino-daemon-impl.test.ts
```diff
@@ -114,18 +114,22 @@ describe('arduino-daemon-impl', () => {
...
-    const instance = await new SilentArduinoDaemonImpl(
+    const { daemon, port } = await new SilentArduinoDaemonImpl(
       'json'
     ).spawnDaemonProcess();
-    expect(instance.port).not.to.be.undefined;
-    expect(instance.port).not.to.be.equal('0');
+
+    expect(port).not.to.be.undefined;
+    expect(port).not.to.be.equal('0');
+    daemon.kill();
```
AlbyIanna: Maybe I'd prefer to put the `new SilentArduinoDaemonImpl('json').spawnDaemonProcess();` in the `beforeEach` and the `daemon.kill()` in the `afterEach`.
@fstasi does it make sense to you?

fstasi: we actually run the `SilentArduinoDaemonImpl` with different input: `json` or `text`.. we can make a wrapper function probably

PR #697, opened 2021-12-16
Serial monitor lines not to wrap lines

fstasi:
## Why
currently, lines plotted to the serial plotter have fixed height, as a this is required for the virtual scroll to work properly and guarantee good performance.

This causes longer lines to wrap and overflow over the line below, causing a glitch in the ui (see picture)

image

## How

Setting the `white-space: nowrap;` property fixes the issue allowing the whole panel to scroll horizontally
image


AlbyIanna (approved):
LGTM

PR #823, opened 2022-02-09
add axe to create accessibility reports

AlbyIanna:
## Why
We want to make the Arduino IDE more accessible.

## How 
We decided to choose [axe](https://www.deque.com/axe/), a tool that does automated accessibility tests at runtime.

## Test instructions
- download the build and launch the executable
- open the developer tools (in the menu **Help > Toggle Developer Tools**)
- open the **Console** tab
- launch the command
```js
runAxe()
```
- you will see some coloured lines spawn in the console reporting accessibility issues, like this:
image
- once you've run this command and until you close the window, as you interact with the application you will keep getting new reports


arduino-ide-extension/package.json
```diff
@@ -93,6 +93,7 @@
     "which": "^1.3.1"
   },
   "devDependencies": {
+    "@axe-core/react": "^4.3.2",
```
kittaakos, line 96: Shouldn't it go under `dependencies` instead of `devDependencies`?

AlbyIanna: Makes sense. Thanks!

AlbyIanna: Thanks for the comment @kittaakos. Basically we're doing this to help the designers testing the app. We'll definitely change it as dev dependency as soon as we find a better solution.

PR #851, opened 2022-02-20
Splashscreen fixes

80avin:
Fixes
1. prevent splashscreen from starting alwaysOnTop
2. prevent pointer events like drag on splash screen

Note: I saw https://github.com/arduino/arduino-ide/pull/717 but the approach is different.

### DIFF
--- a/arduino-ide-extension/src/electron-main/splash/static/splash.html
+++ b/arduino-ide-extension/src/electron-main/splash/static/splash.html
@@ -7,6 +7,7 @@
             width: auto;
             text-align: center;
             padding: 0px;
+            pointer-events: none;
         }
 
         img {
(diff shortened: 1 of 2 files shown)


fstasi:
Hi @80avin ,

thanks for contributing, we definitely need some work on the splash screen.
I tested on mac and I see some behavior we should try to address:

- the splash screen pops up
- it is immediateley covered by the IDE main window (blank)

I'd get rid of the splash screen, we don't need it anymore. Instead, I'd put the splash screen image directly inside the IDE2 window with the loader below. 

This is the mockup our designer @91volt prepared:
Group 254

@80avin let me know if this is something that can be addressed in this PR

Logicer16:
> - the splash screen pops up
> - it is immediateley covered by the IDE main window (blank)

@80avin I know setting `alwaysOnTop` to false would be simpler, but this is why #717 takes a different approach.

I've also noticed that using `pointer-events: none;` still allows the splash screen to be highlighted if it's double clicked.

PR #871, opened 2022-02-28
Go to download page when automatic update fails

AlbyIanna:
### Motivation
Errors happening during the automatic update aren't handled properly. 

### Change description
Whenever an error happens during the automatic update occurs, show a link to the download page to let the user download the IDE by themselves.

### Reviewer checklist

### DIFF
--- a/arduino-ide-extension/src/browser/arduino-frontend-contribution.tsx
+++ b/arduino-ide-extension/src/browser/arduino-frontend-contribution.tsx
@@ -280,7 +280,8 @@ export class ArduinoFrontendContribution
     }
 
     this.updaterService.init(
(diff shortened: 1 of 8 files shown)


fstasi (approved):

per1234 (approved):
I tested under the following conditions:

- Download fails due to a 404
- Download fails due to using the `.zip` Linux build (which doesn't support auto-update)

Previously, either of these conditions resulted in the update download dialog hanging at 0%, with no feedback unless you checked the command line output (and even that wasn't clear in the case of using the `.zip` Linux build).

Now both of these situations are handled gracefully by the IDE.

Thanks Alberto!

PR #893, opened 2022-03-08
Save dialog for closing temporary sketch and unsaved files

msujew:
### Motivation

Closes https://github.com/arduino/arduino-ide/issues/595
Closes https://github.com/arduino/arduino-ide/issues/862

### Change description

When closing the app, it will check whether the current sketch is temporary and ask the user to save it somewhere else.

### Other information

When disabling `autoSave` and editing a file in a temp sketch and trying to close the app, two dialogs will appear. This is an issue in upstream Theia, see https://github.com/eclipse-theia/theia/issues/10860. A PR to fix this is linked.

### Reviewer checklist

* [ ] PR addresses a single concern.
* [ ] The PR has no duplicates (please search among the [Pull Requests](https://github.com/arduino/arduino-ide/pulls) before creating one)
* [ ] PR title and description are properly filled.
* [ ] Docs have been added / updated (for bug fixes / features)

PR #939, opened 2022-04-04
Save all open editors before running `Save As`

msujew:
### Motivation

Closes https://github.com/arduino/arduino-ide/issues/896

### Change description

Invokes the `saveAll` operation of the `ApplicationShell` before saving the sketch to a new location.

### Other information

This saves any unsaved changes to the current sketch as well. Is this alright with the usability concept? Or should the new changes exclusively be saved to the new sketch? (this would be a bit more involved)

### Reviewer checklist

### DIFF
--- a/arduino-ide-extension/src/browser/contributions/save-as-sketch.ts
+++ b/arduino-ide-extension/src/browser/contributions/save-as-sketch.ts
@@ -1,4 +1,4 @@
-import { injectable } from 'inversify';
+import { inject, injectable } from 'inversify';
 import * as remote from '@theia/core/electron-shared/@electron/remote';
 import * as dateFormat from 'dateformat';
 import { ArduinoMenus } from '../menu/arduino-menus';
@@ -11,9 +11,15 @@ import {
   KeybindingRegistry,
 } from './contribution';
 import { nls } from '@theia/core/lib/common';
+import { ApplicationShell } from '@theia/core/lib/browser';
+import { timeout } from '@theia/core/lib/common/promise-util';
 
 @injectable()
 export class SaveAsSketch extends SketchContribution {
+
+  @inject(ApplicationShell)
+  protected readonly applicationShell: ApplicationShell;
+
   registerCommands(registry: CommandRegistry): void {
     registry.registerCommand(SaveAsSketch.Commands.SAVE_AS_SKETCH, {
       execute: (args) => this.saveAs(args),
@@ -87,6 +93,8 @@ export class SaveAsSketch extends SketchContribution {
     if (!destinationUri) {
       return false;
     }
+    await this.applicationShell.saveAll();
+    await timeout(20);
     const workspaceUri = await this.sketchService.copy(sketch, {
       destinationUri,
     });
--- a/arduino-ide-extension/src/browser/contributions/save-as-sketch.ts
+++ b/arduino-ide-extension/src/browser/contributions/save-as-sketch.ts
@@ -11,14 +11,21 @@ import {
   KeybindingRegistry,
 } from './contribution';
 import { nls } from '@theia/core/lib/common';
-import { ApplicationShell } from '@theia/core/lib/browser';
-import { timeout } from '@theia/core/lib/common/promise-util';
+import { ApplicationShell, NavigatableWidget, Saveable } from '@theia/core/lib/browser';
+import { EditorManager } from '@theia/editor/lib/browser';
+import { WindowService } from '@theia/core/lib/browser/window/window-service';
 
 @injectable()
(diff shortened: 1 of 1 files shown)


AlbyIanna (commented):
@msujew Works for me. I have just one doubt: why do we need a timeout? Since `saveAll()` returns a Promise, I would expect not to need any timeout after that.

msujew:
@AlbyIanna Yeah, sadly the internal Theia 1.22 code for `saveAll` looks like this

```ts
Promise.all(widgets, widget => {
  if (Saveable.is(widget)) {
    Saveable.save(widget); // Note the missing "await" here
  }
});
```

Which is why we need to wait. But you're right, overriding the `saveAll` method of the Applicationshell to locally fix this would probably the better idea than waiting. I'll address this.

Note that the saving functionality has been moved into its own service which correctly uses `await` in version 1.23 and onward. We should probably update soon anyway.

AlbyIanna:
Thanks @msujew!

> Perhaps we should try to update to 1.24 soon?

Yes, I think we should.
cc: @fstasi

AlbyIanna:
@ubidefeo oh right that happens to me too with new sketches.
If I follow instructions in #896 👇 


> 1. Select File > Preferences from the Arduino IDE menus.
> 2. Uncheck the box next to "☐ Auto save"
> 3. Click the OK button.
> 4. Select File > New from the Arduino IDE menus.
> ℹ️ The bug is not limited to new sketches. It also occurs with sketches that have been previously saved to a permanent location on the disk.
> 5. Make some change to the sketch.
> 6. ❗ Do not save.
> 7. Select File > Save As... from the Arduino IDE menus.
> 8. Click the Save button.


I get the same buggy behaviour @ubidefeo is describing. 🐛

per1234 (changes requested):
UPDATE: fixed by https://github.com/arduino/arduino-ide/compare/51af711f7b70f6310787360a25294193ddefb280..2543d1b0c81f437e3f8c6dbc60f064bcdb425236

<s>>This saves any unsaved changes to the current sketch as well. Is this alright with the usability concept? Or should the new changes exclusively be saved to the new sketch? 

They must be saved exclusively to the new sketch, just as "Save as" works in every other application ever made.</s>

AlbyIanna:
We are going to merge this together with #893 because we want to solve all the major saving issues together.

AlbyIanna (changes requested):
As @per1234 suggested – we should save the new changes exclusively in the new sketch.

msujew:
@per1234 @AlbyIanna Thanks for the feedback, I've changed the code to only save the changes to the new sketch. Mind retesting?

per1234 (changes requested):
UPDATE: Fixed by https://github.com/arduino/arduino-ide/compare/2543d1b0c81f437e3f8c6dbc60f064bcdb425236..6299e87015c745f7e1588e4b438df893d2c300de

<s>### Describe the problem

When using the build from this PR, performing a "**Save As...**" operation when the sketch has unsaved changes produces an unexpected dialog about quitting:

image

### To reproduce

1. Select **File > Preferences** from the Arduino IDE menus.
1. Uncheck the box next to "**☐ Auto save**"
1. Click the <kbd>OK</kbd> button.
1. Make some change to the sketch (I will refer to this as "Sketch A" from here on).
   ❗ Do not save.
1. Click the <kbd>Save</kbd> button.
    (I will refer to this as "Sketch B" from here on)

🐛 An unexpected and confusing dialog appears:

> Are you sure you want to quit?
>
> Any unsaved changes will not be saved.

### Expected behavior

No quit confirmation dialog on "**Save As...**".

The IDE must be smart enough to understand that, even though it is true that the unsaved changes are not saved to "Sketch A", they are saved to "Sketch B" and it is standard and expected by the user that changes would not be saved to "Sketch A".

The IDE must be smart enough to understand that the closure of "Sketch A" when saving as "Sketch B" is different from a normal close of a window. The replacement of "Sketch A" by "Sketch B" in the application after a "Save As..." operation is simply standard and expected behavior of any application and the user should not be consulted about it or exposed to whatever specific mechanism the IDE is using to accomplish that replacement.

### Arduino IDE version

2.0.0-rc6-snapshot-7ef629a

(the tester build for https://github.com/arduino/arduino-ide/pull/939/commits/2543d1b0c81f437e3f8c6dbc60f064bcdb425236)

### Operating system

Windows 10</s>

msujew:
@per1234 Sorry, small oversight, should be fixed now. That was the underlying Theia `onWillStop` check that popped up.

per1234 (approved):
This solves https://github.com/arduino/arduino-ide/issues/896 for me.
Thanks @msujew!

Issue #866, opened 2022-02-26
Impossible to close launch.json in  Arduino IDE 2 RC3


francip:
**Describe the bug**
Once launch.json is opened through the cog button in the degug view, there is no way to close it. The tab lacks the 'x' button, and any of the shortcuts to close tabs are not working on that tab.
The launch.json tab persist across IDE restart.

**To Reproduce**
1. Open a sketch file
2. Open the Debug view
3. Press the cog button to open launch.json

**Expected behavior**
'x' button is present and allows closing the tab. Keyboard shortcuts for closing tabs are able to close the tab.

**Screenshots**
Screenshot showing the launch.json without the 'x' button and keyboard shortcuts tab with the 'x' button next to it.
![Screenshot 2022-02-25 200644](https://user-images.githubusercontent.com/49422/155828313-d9a31285-e858-44f5-8281-ef00b5e45540.png)

**Desktop (please complete the following information):**
 - OS: Windows 10 Home
 - Version: 2.0.0-rc3 (2021-12-22)

**Additional context**

ubidefeo:
considering the Sketch structure, every file in the Sketch's root should not be closable, but any other file in subdirs or otherwise open should.

`launch.json` is in the `.vscode` folder, hence it should be closeable

PR #946, opened 2022-04-06
Allow to close files in certain folders

msujew:
### Motivation

Closes https://github.com/arduino/arduino-ide/issues/866
Closes https://github.com/arduino/arduino-ide/issues/831

### Change description

When adding a widget to the tracker we first try to see whether it is contained in one of our `ignoreFolders` (`.vscode`, `.theia` and `src` so far). If it is, we allow it to stay closable.

### Reviewer checklist

### DIFF
--- a/arduino-ide-extension/src/browser/theia/core/application-shell.ts
+++ b/arduino-ide-extension/src/browser/theia/core/application-shell.ts
@@ -41,6 +41,14 @@ export class ApplicationShell extends TheiaApplicationShell {
       // Make the editor un-closeable asynchronously.
       this.sketchesServiceClient.currentSketch().then((sketch) => {
         if (sketch) {
+          const ignoreFolders = ['.vscode', '.theia', 'src'];
+          let editorPath = widget.editor.uri;
+          while (editorPath.toString().length > sketch.uri.length) {
+            if (ignoreFolders.includes(editorPath.path.base)) {
+              return;
+            }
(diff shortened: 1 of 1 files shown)


per1234 (commented):
As I see additional folders added selectively to the exclusion, it occurs to me it may be better to go at this from the other direction: by simply making the root of the sketch folder the "non-closable" area, with any files from subfolders always being closable.

That will provide symmetry with the sketch files that are opened by default and avoid the poor user experience of being able to open an additional file via the Sketchbook view (when "**☐ Show files inside Sketches**" is enabled), but then not being allowed to close it again.

For example, the Arduino Sketch Specification provides [a `data` subfolder](https://arduino.github.io/arduino-cli/dev/sketch-specification/#data-subfolder) "to add additional files to the sketch, which will not be compiled". Since these will not be part of the sketch program by definition, there is no reason we should force the user to keep the tab open.

AlbyIanna:
@per1234 that makes sense. @msujew do you think you can tackle this as Per suggested?

msujew:
@per1234 Thanks for the input, I've updated the code to only prevent file from closing which are direct children of the sketch directory. As far as I could see in the sketch spec this seems to be the intended behavior. Or does it only apply to `*.ino` files?

kittaakos:
Follow up: https://github.com/arduino/arduino-ide/issues/977

PR #957, opened 2022-04-15
Set the current language on the localization provider

AlbyIanna:
### Motivation
The English language gets forgotten when changing the language.
Solves #956 

### Change description
Set the language to the localization provider

### Reviewer checklist

### DIFF
--- a/arduino-ide-extension/src/browser/dialogs/settings/settings.tsx
+++ b/arduino-ide-extension/src/browser/dialogs/settings/settings.tsx
@@ -278,9 +278,9 @@ export class SettingsService {
     // after saving all the settings, if we need to change the language we need to perform a reload
     // Only reload if the language differs from the current locale. `nls.locale === undefined` signals english as well
     if (
-      currentLanguage !== nls.locale &&
-      !(currentLanguage === 'en' && nls.locale === undefined)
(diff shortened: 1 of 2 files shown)


arduino-ide-extension/src/browser/dialogs/settings/settings.tsx
```diff
@@ -278,9 +278,9 @@ export class SettingsService {
...
     // Only reload if the language differs from the current locale. `nls.locale === undefined` signals english as well
     if (
-      currentLanguage !== nls.locale &&
-      !(currentLanguage === 'en' && nls.locale === undefined)
+      currentLanguage !== (await this.localizationProvider.getCurrentLanguage())
     ) {
+      await this.localizationProvider.setCurrentLanguage(currentLanguage);
       if (currentLanguage === 'en') {
         window.localStorage.removeItem(nls.localeId);
       } else {
```
AlbyIanna, line 286: @kittaakos do you think it's necessary to save the locale id to the local storage like this in order for the localization to work?

I thought that setting the current language to the localization provider was enough, so I've tried removing this code, but then the localization didn't work correctly, therefore it appears it is not enough.

per1234 (approved):
This solves https://github.com/arduino/arduino-ide/issues/956 for me.
Thanks Alberto!

Issue #944, opened 2022-04-05
Authentication sessions are not persistent


fstasi:
### Describe the problem

Authentication sessions with the arduino cloud are not persistent between application lifecycles

### To reproduce

1. open the IDE2
2. open the remote sketchbook
3. login using you arduino credentials
4. close the application
5. re open the application
6. open the remote sketchbook
7. ❗️verify that login is needed again

### Expected behavior

I expect the login to be persistent between different IDE2 runs.

### Arduino IDE version

2.0.0-rc5

### Operating system

macOS

### Operating system version

12.3.1

### Issue checklist

- [X] I searched for previous reports in [the issue tracker](https://github.com/arduino/arduino-ide/issues?q=)
- [X] I verified the problem still occurs when using the latest [nightly build](https://github.com/arduino/arduino-ide#nightly-builds)
- [X] My report contains all necessary details

PR #992, opened 2022-05-18
#944: Fixed auth. sessions not persistent

davegarthsimpson:
### Motivation
To ensure login sessions to Arduino Cloud are persistent when closing and reopening the application.

### Issue
We were initialising our backend auth service and performing token refreshes without the required "authOptions" properties. These properties are set on the frontend after the backend service is initialised.

### Change description
We now perform the auth service initialisation and token refreshes on the frontend after the required "authOptions" properties are set.

> Removed initialisation from the backend service;
> Created a method on the backend service to initialise the auth service from the frontend (and added the method to the related interface);
> Created an "initialised" variable in the backend auth service to ensure initialisation does not run more than once (incase of multiple frontend instances);
> Invoked initialisation on the frontend service after setting required "authOptions" properties;

### Reviewer checklist

### DIFF
--- a/arduino-ide-extension/src/browser/auth/authentication-client-service.ts
+++ b/arduino-ide-extension/src/browser/auth/authentication-client-service.ts
@@ -49,7 +49,10 @@ export class AuthenticationClientService
     this.service
       .session()
       .then((session) => this.notifySessionDidChange(session));
+
     this.setOptions();
+    this.service.initAuthSession()
+
     this.arduinoPreferences.onPreferenceChanged((event) => {
       if (event.preferenceName.startsWith('arduino.auth.')) {
         this.setOptions();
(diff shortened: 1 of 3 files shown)


ubidefeo (approved):
Works as expected
LGTM

arduino-ide-extension/src/browser/auth/authentication-client-service.ts
```diff
@@ -49,7 +49,10 @@ export class AuthenticationClientService
     this.service
       .session()
       .then((session) => this.notifySessionDidChange(session));
+
     this.setOptions();
+    this.service.initAuthSession()
```
kittaakos, line 54: Note, above this `initAuthSession` call, the `setOptions` is a void, hence notification. It's a "fire and forget" event, so by the time you call `initAutSession` it is not ensured that the options were set or arrived at all.

In the long run, we have to change all similar calls from `foo(arg: any): void` to `foo(arg: any): Promise<void>`, and change the code from this

```ts
this.setOptions();
this.service.initAuthSession()
```

to this:

```ts
```ts
await this.setOptions();
this.service.initAuthSession()
```

//CC @fstasi, @AlbyIanna, @davegarthsimpson, and @francescospissu

PR #1058, opened 2022-06-13
Output panel optimisation

davegarthsimpson:
### Motivation
Output panel prints slower than the CLI and on windows is out of sync with the upload operation.

### Change description
Adds a simple buffer mechanism for the related stream from the CLI.

### Other information
Optimisations on CLI side should also complement the resolution of this issue.

### Reviewer checklist

### DIFF
--- a/arduino-ide-extension/src/node/core-service-impl.ts
+++ b/arduino-ide-extension/src/node/core-service-impl.ts
@@ -78,7 +78,7 @@ export class CoreServiceImpl extends CoreClientAware implements CoreService {
     const result = client.compile(compileReq);
 
     const compileBuffer = new OutputPanelBufferProvider(
-      this.flushOutputPanelMessages,
+      this.flushOutputPanelMessages.bind(this),
       this.FLUSH_OUTPUT_MESSAGES_TIMEOUT_MS
     );
     try {
@@ -185,7 +185,7 @@ export class CoreServiceImpl extends CoreClientAware implements CoreService {
     const result = responseHandler(client, req);
 
     const uploadBuffer = new OutputPanelBufferProvider(
-      this.flushOutputPanelMessages,
+      this.flushOutputPanelMessages.bind(this),
       this.FLUSH_OUTPUT_MESSAGES_TIMEOUT_MS
     );
     try {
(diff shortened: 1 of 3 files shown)


arduino-ide-extension/src/node/utils/simple-buffer.ts
```diff
@@ -0,0 +1,31 @@
+export class SimpleBuffer {
+  chunks: Uint8Array[] = [];
```
kittaakos, line 2: Missing `private`?

arduino-ide-extension/src/node/utils/simple-buffer.ts
```diff
@@ -0,0 +1,31 @@
...
+        onFlush(chunkString);
+      }
+    }, flushTimeout);
+  }
+
+  public addChunk(chunk: Uint8Array): void {
+    this.chunks.push(chunk);
+  }
+
+  private clearChunks() {
```
kittaakos, line 21: Missing return type?

arduino-ide-extension/src/node/core-service-impl.ts
```diff
@@ -39,6 +40,8 @@ export class CoreServiceImpl extends CoreClientAware implements CoreService {
 
   protected uploading = false;
 
+  private FLUSH_OUTPUT_MESSAGES_TIMEOUT_MS = 32;
```
kittaakos, line 43: Missing readonly? If this is a static constant, it should be `static`. Or please move out from the class to the module scope and use `const`. Having a variable in the class with `SUCH_NAMING` is confusing. Thank you!

per1234 (approved):
I tested with the build from https://github.com/arduino/arduino-ide/pull/1058/commits/085f97e9989ea891c0581b822c261c3fa02179a7 and this fixes both https://github.com/arduino/arduino-ide/issues/68 and https://github.com/arduino/arduino-ide/issues/405 for me.

In order to do a more thorough test of https://github.com/arduino/arduino-ide/issues/68, I replaced the standard upload tool command with a script that prints 10000 lines of text. That results in a large delay between the completion of the output process and the completion of printing to the Output panel when using the latest build from the `main` branch. Using the build from this PR, the output printing completes at the same time as the output process.

Thanks Dave!

PR #1123, opened 2022-06-28
change output buffer to setTimeout instead of setInterval

davegarthsimpson:
### Motivation
To make the buffer mechanism of the output panel more robust.

### Change description
Uses a setTimeout rather that setInterval in our buffer, this ensures that flush calls do not interfere with each other. On disposing our buffer we set a boolean flag to prevent additional setTimeout callbacks from being created, if there is a setTimeout callback in progress when the buffer is "disposed" it will complete before running flush a final time to be exhaustive.

gRPC errors are also added to the buffer, to ensure proper ordering in the output panel.

### Reviewer checklist


arduino-ide-extension/src/node/utils/simple-buffer.ts
```diff
@@ -33,11 +35,39 @@ export class SimpleBuffer implements Disposable {
...
+      this.flushTimeout = undefined;
+    }
+  }
+
   dispose(): void {
-    this.flush();
-    clearInterval(this.flushInterval);
-    this.clearChunks();
-    this.flushInterval = undefined;
+    this.disposed = true;
```
davegarthsimpson, line 70: alternatively we could use promises, rather than a flag, awaiting here and then performing a final flush

arduino-ide-extension/src/node/core-service-impl.ts
```diff
@@ -76,15 +76,19 @@ export class CoreServiceImpl extends CoreClientAware implements CoreService {
...
             );
+            handler.addChunk(chunk, OutputMessage.Severity.Error);
+
             reject(CoreError.VerifyFailed(message, compilerErrors));
           }
         })
-        .on('end', resolve);
-    }).finally(() => handler.dispose());
+        .on('end', () => {
+          handler.dispose();
```
davegarthsimpson, line 88: the `on('end'` callback will always be called even if there is an error. It fires after the `on('error'` callback, and signifies there are no more messages to be receive, hence doing `handler.dispose` on `.finally` is incorrect. As `.on('error'` will reject the promise, we may get to the `.finally` callback before the `on('end'` callback is called, in this case messages recieved after the error would not be captured

arduino-ide-extension/src/node/utils/simple-buffer.ts
kittaakos: Thanks for bringing it up. The flag is looking good. No need to over-engineer it.

kittaakos:
I proposed a few changes [here](https://github.com/arduino/arduino-ide/compare/use-timeout-in-output-buffer...kittaakos:arduino-ide:buffer2). What do you think?

arduino-ide-extension/src/node/utils/buffers.ts
```diff
@@ -69,4 +77,4 @@ namespace Chunks {
       ])
     );
   }
-}
+}
```
kittaakos, line 80: Missing newline.

kittaakos:
@fstasi, please review. Thank you!

AlbyIanna (approved):
LGTM

davegarthsimpson (commented):
note for traceability: changes in [e5f922a](https://github.com/arduino/arduino-ide/pull/1123/commits/e5f922abfe2b2e2ace3f813d40e560274d4ae892) were implemented by @kittaakos, then accepted and copied here by @davegarthsimpson.

kittaakos:
The PR looks good to me. Please proceed with the merge if this is ready. Thank you!

PR #1151, opened 2022-07-06
Window min size

AlbyIanna:
### Motivation
The current window minimum size is too low, we want to increase it.

### Change description
- [set the minWidth and minHeight](https://github.com/arduino/arduino-ide/compare/window-min-size?expand=1#diff-54951a7f6fc1d8e5bca375af44a6874aa403c6ab0942031b7126743fc7d5ab66R366-R368) when opening a new window (note that I also set `useContentSize` to `true` to ensure the top bar of the OS is not included in the minHeight
- changes the style of the dialogs to ensure the UI doesn't glitch when the user resizes the window  

### Reviewer checklist

### DIFF
--- a/arduino-ide-extension/src/electron-main/theia/electron-main-application.ts
+++ b/arduino-ide-extension/src/electron-main/theia/electron-main-application.ts
@@ -339,6 +339,9 @@ export class ElectronMainApplication extends TheiaElectronMainApplication {
       options.webPreferences = {};
     }
     options.webPreferences.v8CacheOptions = 'bypassHeatCheck'; // TODO: verify this. VS Code use this V8 option.
+    options.useContentSize = true;
+    options.minWidth = 680;
+    options.minHeight = 565;
     return options;
   }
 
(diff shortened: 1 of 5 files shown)


arduino-ide-extension/src/electron-main/theia/electron-main-application.ts
```diff
@@ -339,6 +363,9 @@ export class ElectronMainApplication extends TheiaElectronMainApplication {
       options.webPreferences = {};
     }
     options.webPreferences.v8CacheOptions = 'bypassHeatCheck'; // TODO: verify this. VS Code use this V8 option.
+    options.useContentSize = true;
+    options.minWidth = 680;
+    options.minHeight = 565;
```
kittaakos, line 368: Could you override `getDefaultTheiaWindowOptions` and do the customization there?

AlbyIanna: Yes I could, but can I ask you why do you think it's better to set this option inside `getDefaultTheiaWindowOptions` rather than `getDefaultOptions`?

My rationale to put it inside `getDefaultOptions` is that Theia sets `minWidth` and `minHeight` there.

kittaakos: You're absolutely right. Thanks for asking for clarification.

fstasi (approved):
well done, works as expected

PR #1166, opened 2022-07-11
Add language packs

AlbyIanna:
### Motivation
As on [Transifex](https://www.transifex.com/arduino-1/ide2/) we are getting more and more strings translated we want to add support for new languages.

### Reviewer checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -80,10 +80,15 @@
     "vscode-builtin-json-language-features": "https://open-vsx.org/api/vscode/json-language-features/1.46.1/file/vscode.json-language-features-1.46.1.vsix",
     "cortex-debug": "https://open-vsx.org/api/marus25/cortex-debug/0.3.10/file/marus25.cortex-debug-0.3.10.vsix",
     "vscode-language-pack-nl": "https://open-vsx.org/api/MS-CEINTL/vscode-language-pack-nl/1.48.3/file/MS-CEINTL.vscode-language-pack-nl-1.48.3.vsix",
-    "vscode-language-pack-fr": "https://open-vsx.org/api/MS-CEINTL/vscode-language-pack-fr/1.53.2/file/MS-CEINTL.vscode-language-pack-fr-1.53.2.vsix",
-    "vscode-language-pack-zh-hans": "https://open-vsx.org/api/MS-CEINTL/vscode-language-pack-zh-hans/1.53.2/file/MS-CEINTL.vscode-language-pack-zh-hans-1.53.2.vsix",
-    "vscode-language-pack-de": "https://open-vsx.org/api/MS-CEINTL/vscode-language-pack-de/1.53.2/file/MS-CEINTL.vscode-language-pack-de-1.53.2.vsix",
-    "vscode-language-pack-ja": "https://open-vsx.org/api/MS-CEINTL/vscode-language-pack-ja/1.53.2/file/MS-CEINTL.vscode-language-pack-ja-1.53.2.vsix",
-    "vscode-language-pack-tr": "https://open-vsx.org/api/MS-CEINTL/vscode-language-pack-tr/1.53.2/file/MS-CEINTL.vscode-language-pack-tr-1.53.2.vsix"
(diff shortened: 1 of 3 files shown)


per1234 (changes requested):
It is necessary to also add these to the `theiaPlugins` key of `electron/build/template-package.json`:

https://github.com/arduino/arduino-ide/blob/7fed8febf156c4cec25f9053efb065575df7a1ee/electron/build/template-package.json#L142-L154

Otherwise they will not be in the distribution package:

image

package.json
```diff
@@ -80,10 +80,15 @@
...
+    "vscode-language-pack-de": "https://open-vsx.org/api/MS-CEINTL/vscode-language-pack-de/1.69.0/file/MS-CEINTL.vscode-language-pack-de-1.69.0.vsix",
+    "vscode-language-pack-ja": "https://open-vsx.org/api/MS-CEINTL/vscode-language-pack-ja/1.69.0/file/MS-CEINTL.vscode-language-pack-ja-1.69.0.vsix",
+    "vscode-language-pack-tr": "https://open-vsx.org/api/MS-CEINTL/vscode-language-pack-tr/1.69.0/file/MS-CEINTL.vscode-language-pack-tr-1.69.0.vsix",
+    "vscode-language-pack-it": "https://open-vsx.org/api/MS-CEINTL/vscode-language-pack-it/1.69.0/file/MS-CEINTL.vscode-language-pack-it-1.69.0.vsix",
+    "vscode-language-pack-ru":"https://open-vsx.org/api/MS-CEINTL/vscode-language-pack-ru/1.69.0/file/MS-CEINTL.vscode-language-pack-ru-1.69.0.vsix",
+    "vscode-language-pack-es": "https://open-vsx.org/api/MS-CEINTL/vscode-language-pack-es/1.69.0/file/MS-CEINTL.vscode-language-pack-es-1.69.0.vsix",
+    "vscode-language-pack-pt-BR": "https://open-vsx.org/api/MS-CEINTL/vscode-language-pack-pt-BR/1.69.0/file/MS-CEINTL.vscode-language-pack-pt-BR-1.69.0.vsix",
```
per1234, line 91: https://github.com/arduino/arduino-ide/blob/55e83a89d82903b2f21ce47d5fd5d4d3f842480c/arduino-ide-extension/src/node/arduino-localization-contribution.ts#L32-L35

I found that the `locale` argument must match the `contributes.localizations.languageId` value from the language pack in order for the Arduino-specific localizations to take effect. 

This was the only value that did not match with the language packs installed via this PR.

https://github.com/microsoft/vscode-loc/blob/a0cfb4e27f712122066e6f994ef1403e807389bf/i18n/vscode-language-pack-pt-BR/package.json#L22

```json
"languageId": "pt-br",
```

The Arduino IDE after selecting **File > Preferences > Language > português (Brasil)** when using the build from https://github.com/arduino/arduino-ide/commit/55e83a89d82903b2f21ce47d5fd5d4d3f842480c:

image


The Arduino IDE after changing the the `locale` argument to `pt-br`:

image

However, I'm wondering about whether it is appropriate to use the `pt-br` language pack with the Arduino Portuguese localization data. I don't see a language pack for `pt` [in the Open VSX Registry](https://open-vsx.org/?search=&category=Language%20Packs).

per1234 (changes requested):

package.json
AlbyIanna: Nice one, thanks. I don't really know what to do here. The reason I chose the Brasilian pack is that it was the only Portuguese pack, but I'm having second thoughts on this choice.

AlbyIanna: I've removed the Portuguese language pack _for now_. I'll re-add it ASAP when we find out how to handle this.

per1234 (approved):
Very exciting to see the list of available languages grow.

Thanks Alberto!

 Thanks also to the translators for helping to make the Arduino IDE accessible to the entire world!

package.json
AlbyIanna: After talking about this with our localization specialist, I learned that for now want to avoid splitting dialects (like `pt` and `pt-br`) so as not to disperse the forces of the community. Since the people contributing to the Portuguese translation on Transifex are a mix of Portuguese and Brazilians, we decided to use the `pt-br` language pack.

PR #1168, opened 2022-07-11
arm64

koendv:
### Motivation
Build arduino-ide for arm64

### Change description
Adds build documentation and build script

### Other information
These are instructions on compiling [arduino-ide 2](https://github.com/arduino/arduino-ide) for arm64 using self-hosted runners. Compiling can be done on raspberry pi os 64-bit or ubuntu arm64. The compiled binaries run on raspberry pi os 64-bit and ubuntu arm64.

Compiling is done using github actions, the same build process as on X86.

Because github does not offer linux arm64 runners, compilation is done using a self-hosted runner inside a docker on a raspberry pi. 

Because github does not recommend running self-hosted runners on public repositories, the self-hosted runners are started from a fork of the arduino-ide repository, with public PR switched off on the fork.

A definitive solution would be github offering public linux/arm64 runners. 

### Reviewer checklist

### DIFF
--- a/docs/self-hosted-runner/SelfHostedRunner.md
+++ b/docs/self-hosted-runner/SelfHostedRunner.md
@@ -1,0 +1,143 @@
+# arduino-ide 2.0.0 for Raspberry Pi
+
+These are instructions on compiling [arduino-ide 2](https://github.com/arduino/arduino-ide) for arm64 using self-hosted runners. Compiling can be done on raspberry pi os 64-bit or ubuntu arm64. The compiled binaries run on raspberry pi os 64-bit and ubuntu arm64.
+
+Compiling is done using github actions, the same build process as on X86.
+
+Because github does not offer linux arm64 runners, compilation is done using a self-hosted runner on a raspberry pi. 
+
+Because github does not recommend running self-hosted runners on public repositories, the arduino-ide repository is forked first, and public PR are switched off on the fork.
+
+For repeatability and security, compilation is done inside a docker.
+
+## fork arduino-ide
+
+Set up a copy of the arduino-ide github.
+
+- in github, fork [arduino/arduino-ide](https://github.com/arduino/arduino-ide)
+- patch your fork with the file _self_hosted_runner.patch_. The patch adds a new build target "self-hosted" to arduino-ide, next to the existing Windows, Ubuntu and MacOS.
+- for cosmetics, set a tag to show the release name
+
+```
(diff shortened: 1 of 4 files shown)


docs/self-hosted-runner/SelfHostedRunner.md
```diff
@@ -0,0 +1,143 @@
...
+
+These are instructions on compiling [arduino-ide 2](https://github.com/arduino/arduino-ide) for arm64 using self-hosted runners. Compiling can be done on raspberry pi os 64-bit or ubuntu arm64. The compiled binaries run on raspberry pi os 64-bit and ubuntu arm64.
+
+Compiling is done using github actions, the same build process as on X86.
+
+Because github does not offer linux arm64 runners, compilation is done using a self-hosted runner on a raspberry pi. 
+
+Because github does not recommend running self-hosted runners on public repositories, the arduino-ide repository is forked first, and public PR are switched off on the fork.
+
+For repeatability and security, compilation is done inside a docker.
```
umbynos, line 11: This approach could be enhanced, IMHO. Having a dedicated runner is a problem:
- we have to manually operate on it and update it regularly 
- it's a security risk
- I don't think that having a forked private version only for the self-hosted runner will pay off in the future
- 
Maybe using this amazing piece of technology would be better:
https://github.com/docker/setup-qemu-action
Basically, this allows to run docker containers built for another architecture directly inside the runners provided by GitHub. Like this one for example: https://hub.docker.com/r/amd64/ubuntu/

We use the same approach to build [imgtool](https://github.com/mcu-tools/mcuboot/tree/main/scripts)
https://github.com/arduino/imgtool-packing/blob/bcb0ac50d94633784c51a0833fe2346e8cf8b2aa/.github/workflows/release.yml#L118-L187

docs/self-hosted-runner/self_hosted_runner.patch
```diff
@@ -0,0 +1,70 @@
+diff --git a/.github/workflows/build.yml b/.github/workflows/build.yml
```
umbynos, line 1: Instead of committing a patch, would be better to apply this directly and commit the result

umbynos (changes requested):
Thanks a lot for the contribution!
I think the process could be enhanced a bit.
Unfortunately, using self-hosted runners is currently not feasible for us... Maybe a good idea would be to use https://github.com/docker/setup-qemu-action and emulation.
Let me know If you need help from my side :smiley:

Issue #998, opened 2022-05-20
[UI] The address of the board is off from the board selection dropdown


kittaakos:
### Describe the problem

When the address of the connected board is "too long", it can go off from the board selector dropdown widget.
<img width="259" alt="Screen Shot 2022-05-20 at 10 58 48" src="https://user-images.githubusercontent.com/1405703/169493766-4b7c6315-5875-4c7e-9c2a-c386663531c1.png">

### To reproduce

 - Open the IDE and connect a board that has a "long enough" address,
 - Click on the dropdown.

### Expected behavior

The address of the board is not off from the dropdown widget.

### Arduino IDE version

`HEAD` of the `main` branch (b407d0aee079a9a47c94c06c00358d74e8870a10)

### Operating system

macOS

### Operating system version

12.3.1

### Additional context

I had 100% UI scale settings.

### Issue checklist

- [X] I searched for previous reports in [the issue tracker](https://github.com/arduino/arduino-ide/issues?q=)
- [X] I verified the problem still occurs when using the latest [nightly build](https://github.com/arduino/arduino-ide#nightly-builds)
- [X] My report contains all necessary details

PR #1216, opened 2022-07-19
Prevent board selector item labels to overflow

AlbyIanna:
### Motivation
As stated in #998, the items of the Board Selector dropdown may overflow in an ugly way if the label is too long. 

### Change description
Adding `overflow: hidden` to parent and ellipsis to children.
I also added a tooltip to show the complete board/port combination, so that a user is able to see when hovering over the item.

### Reviewer checklist

### DIFF
--- a/arduino-ide-extension/src/browser/boards/boards-toolbar-item.tsx
+++ b/arduino-ide-extension/src/browser/boards/boards-toolbar-item.tsx
@@ -130,11 +130,14 @@ export class BoardsDropDown extends React.Component<BoardsDropDown.Props> {
             protocolIcon
           )}
         />
-        <div className="arduino-boards-dropdown-item--label">
-          <div className="arduino-boards-dropdown-item--board-label">
+        <div
+          className="arduino-boards-dropdown-item--label"
+          title={`${boardLabel}\n${port.address}`}
+        >
+          <div className="arduino-boards-dropdown-item--board-label noWrapInfo noselect">
             {boardLabel}
           </div>
-          <div className="arduino-boards-dropdown-item--port-label">
+          <div className="arduino-boards-dropdown-item--port-label noWrapInfo noselect">
             {port.address}
           </div>
         </div>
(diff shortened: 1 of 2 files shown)


kittaakos:
I could not verify it; non of my boards pop up with such a long address. I propose getting back to this after RC-9.

91volt (approved):

AlbyIanna:
@91volt did the functional review.
@kittaakos does the code look good to you?
 
If you want to try this PR anyway, you can just open the Board Selector, click on 'Select other board and port' and select any board with a long name, like 'Arduino Circuit Playground Express'.

kittaakos:
> f you want to try this PR anyway, you can just open the Board Selector, click on 'Select other board and port' and select any board with a long name, like 'Arduino Circuit Playground Express'.

Smart 👍 


Trying it with and without the PR, I do not think overflow is better here.

Without the changes (the user can see the full name of the board):
Screen Shot 2022-07-29 at 17 41 17


With your changes (it's trimmed and only the ellipses are visible):
Screen Shot 2022-07-29 at 17 40 45


From the original bug report:

> When the address of the connected board is "too long", it can go off from the board selector dropdown widget.
> Screen Shot 2022-05-20 at 10 58 48

See that the numbers are off from the dropdown. It's not a problem with  'Arduino Circuit Playground Express'. It would be great to verify it with a real example.

AlbyIanna:
> With your changes (it's trimmed and only the ellipses are visible):

This was actually expected by design, the dropdown's items should have a fixed height.
@91volt can you confirm this?

@kittaakos consider that I've also added a tooltip that shows the whole board and port when hovering on an item
image


> It would be great to verify it with a real example.

If you want you can edit the board name or the port by inspecting and editing the HTML in the dev tools, the result should be the same.

91volt:
> > With your changes (it's trimmed and only the ellipses are visible):
> 
> This was actually expected by design, the dropdown's items should have a fixed height. @91volt can you confirm this?
> 
> @kittaakos consider that I've also added a tooltip that shows the whole board and port when hovering on an item image
> 
> > It would be great to verify it with a real example.
> 
> If you want you can edit the board name or the port by inspecting and editing the HTML in the dev tools, the result should be the same.

Yep I confirm that.

kittaakos (approved):

PR #1237, opened 2022-07-21
#374: ensure compile verbose pref is included on upload

davegarthsimpson:
### Motivation
Verbose output preferences reported as incorrect during upload actions.

### Change description
Ensure compile "verbose" preference is considered when performing compile step prior to upload.

### Other information
Upload "verbose" preference was being used for both compile and upload steps of the "upload operation".

### Reviewer checklist

### DIFF
--- a/arduino-ide-extension/src/browser/contributions/upload-sketch.ts
+++ b/arduino-ide-extension/src/browser/contributions/upload-sketch.ts
@@ -217,6 +217,7 @@ export class UploadSketch extends CoreServiceContribution {
         verbose,
         sourceOverride,
         optimizeForDebug,
+        compileVerbose,
       ] = await Promise.all([
         this.boardsDataStore.appendConfigToFqbn(
           boardsConfig.selectedBoard?.fqbn
@@ -228,7 +229,12 @@ export class UploadSketch extends CoreServiceContribution {
         this.commandService.executeCommand<boolean>(
           'arduino-is-optimize-for-debug'
         ),
+        this.preferences.get('arduino.compile.verbose'),
       ]);
+
+      const compileStepOptions: Partial<CoreService.Compile.Options> = {
+        verbose: compileVerbose,
+      };
 
       const board = {
         ...boardsConfig.selectedBoard,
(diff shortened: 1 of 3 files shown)


per1234 (approved):
This fixes https://github.com/arduino/arduino-ide/issues/374 for me.

Thanks Dave!

kittaakos:
Can we clean up the typings and the extra options argument for better APIs?

[Here](https://github.com/arduino/arduino-ide/compare/374-verbose-output-pref-not-applied-correctly...kittaakos:better-typings) is the TypeScript way of redefining the `verbose` `boolean` type to `{ compile: boolean; upload: boolean }` when uploading (using a programmer).

See the docs about `Omit` [here](https://www.typescriptlang.org/docs/handbook/utility-types.html#omittype-keys).

davegarthsimpson:
> Here

I don't mind changing it to this, I favoured the separation of compile and upload options to be more explicit, and maybe make things easier in case we need to pass additional compile specific options into the upload op in the future. What do you think @kittaakos?

kittaakos:
> I favoured the separation of compile and upload options to be more explicit

I do not want to introduce additional args in the service calls for a bug fix. First, the JS way is to provide one option, not multiple ones. Second, if the second arg is partial, a dev can specify additional options that are not used, only the `verbose`. What would be the expected behavior of this call with your API proposal:

```ts
coreService.upload({ board: { fqbn: 'a' }, verbose: true }, { board: { fqbn: 'b' }, verbose: false });
```

davegarthsimpson:
> > I favoured the separation of compile and upload options to be more explicit
> 
> I do not want to introduce additional args in the service calls for a bug fix. First, the JS way is to provide one option, not multiple ones. Second, if the second arg is partial, a dev can specify additional options that are not used, only the `verbose`. What would be the expected behavior of this call with your API proposal:
> 
> ```ts
> coreService.upload({ board: { fqbn: 'a' }, verbose: true }, { board: { fqbn: 'b' }, verbose: false });
> ```

I appreciate that making a service call, usually you'd only expect one options object, but my thinking was: the `doUpload` call doesn't do just one thing, it also calls another service `compile`, so maybe there is value in being explicit about what options are use for each step.

No problem though, if the priority is to maintain the APIs' style + if you see no value in separating it out let's go for your suggested approach.

davegarthsimpson:
> when uploading (using a programmer).

what do you mean here by "using a programmer"? [better verbose typings](https://github.com/arduino/arduino-ide/pull/1237/commits/93a8bd943066474b2863928194d67137debdd7a2) also works when `usingProgrammer` is false no?

fstasi (approved):
lgtm

Issue #1223, opened 2022-07-19
IDE2 shows correctly picks the theme based on the OS' theme, but it's incorrectly shown in the `Settings`


kittaakos:
### Describe the problem

When I start the IDE2 for the first time and have a dark theme enabled, the IDE2 correctly comes up with the `Dark (Arduino)` theme,  but the `Light (Arduino)` theme is shown in the _Settings_.

### To reproduce

 - Mimic a first start, (delete `~/.arduinoIDE/settings.json` and execute `localStorage.clear()` or equivalent)
 - Have an OS dark theme,
 - Start the IDE2, (IDE2 starts with the dark theme ✅ )
 - Open the `Settings`, it shows the light theme 🐛 

<img width="866" alt="Screen Shot 2022-07-19 at 15 48 58" src="https://user-images.githubusercontent.com/1405703/179767224-5980470b-080c-434d-a2d1-064f5cc1b7e0.png">

### Expected behavior

The dark theme is selected.

### Arduino IDE version

7680194feb4294cca72fd820617113d316d579b3

### Operating system

macOS

### Operating system version

12.3.1

### Issue checklist

- [X] I searched for previous reports in [the issue tracker](https://github.com/arduino/arduino-ide/issues?q=)
- [X] I verified the problem still occurs when using the latest [nightly build](https://github.com/arduino/arduino-ide#nightly-builds)
- [X] My report contains all necessary details

kittaakos:
Here is another problem, after the first start, open the preferences (it shows the light theme as described above), increase the UI scale, decrease it, press OK, IDE2 theme changes to light.

PR #1238, opened 2022-07-21
#1223: use theme service on settings load

davegarthsimpson:
### Motivation
Incorrect theme shown on first start in settings.

### Change description
Use `matchMedia` method to determine initial UI dropdown value. Patch the app config provider to dispatch based on the OS' theme.

### Reviewer checklist

### DIFF
--- a/arduino-ide-extension/src/browser/dialogs/settings/settings.ts
+++ b/arduino-ide-extension/src/browser/dialogs/settings/settings.ts
@@ -122,7 +122,6 @@ export class SettingsService {
       languages,
       currentLanguage,
       editorFontSize,
-      themeId,
       autoSave,
       quickSuggestions,
       autoScaleInterface,
@@ -137,10 +136,6 @@ export class SettingsService {
       ['en', ...(await this.localizationProvider.getAvailableLanguages())],
       this.localizationProvider.getCurrentLanguage(),
       this.preferenceService.get<number>(FONT_SIZE_SETTING, 12),
-      this.preferenceService.get<string>(
-        'workbench.colorTheme',
-        'arduino-theme'
-      ),
       this.preferenceService.get<Settings.AutoSave>(
         AUTO_SAVE_SETTING,
         Settings.AutoSave.DEFAULT_ON
@@ -165,7 +160,7 @@ export class SettingsService {
     const sketchbookPath = await this.fileService.fsPath(new URI(sketchDirUri));
     return {
       editorFontSize,
-      themeId,
+      themeId: ThemeService.get().getCurrentTheme().id,
       languages,
       currentLanguage,
       autoSave,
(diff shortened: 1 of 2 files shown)


arduino-ide-extension/src/browser/dialogs/settings/settings.ts
```diff
@@ -137,10 +136,6 @@ export class SettingsService {
       ['en', ...(await this.localizationProvider.getAvailableLanguages())],
       this.localizationProvider.getCurrentLanguage(),
       this.preferenceService.get<number>(FONT_SIZE_SETTING, 12),
-      this.preferenceService.get<string>(
-        'workbench.colorTheme',
-        'arduino-theme'
```
kittaakos, line 142: What about generalizing [this](https://github.com/arduino/arduino-ide/blob/5087ff08f216773c35de7ecc8300b771eafea4a9/electron/build/patch/frontend/index.js#L17-L22) logic, and replacing the hard-coded `'arduino-theme'` default value with it?

Something like this:
```ts
this.preferenceService.get<string>(
  'workbench.colorTheme',
  window.matchMedia &&
    window.matchMedia('(prefers-color-scheme: dark)').matches
    ? 'arduino-theme-dark'
    : 'arduino-theme'
),
```

davegarthsimpson: For me it seemed logical to use the `ThemeService` as we go on to use `ThemeService.get` in the JSX of the `settings-component.tsx`.

I'm probably missing some info. too... on my mac, on first start with a dark themed os, the Theia (Dark) theme is loaded not the Arduino one, is this unexpected?

Are you suggesting wrapping that logic? or just reusing it as above?

kittaakos: > For me it seemed logical to use 

The method is `loadSettings`; but it does not load the theme. It uses the value from the `ThemeService`. All other settings are read from the `settings.json` or the `arduino-cli.yaml`. Your implementation relies on how the `workbench.colorTheme` preference value changes are picked up by the `ThemeService` form Theia, but it can change. I am proposing this change to have complete control of our code.

> Are you suggesting wrapping this logic? or just reusing it as above?

Reusing the same logic would be the best, but since the theme hack will be removed, you can reuse the code here.

davegarthsimpson: understood, lets go for that then

kittaakos:
My latest commit (387e52b) patches the app config provider. The default value from Theia is based on the app config. See [here](https://github.com/eclipse-theia/theia/blob/896c343f59281e7710388969b2ae89f08b2130e7/packages/core/src/browser/core-preferences.ts#L151-L155). It always picked up the light theme, now it should dispatch based on the OS' theme.

kittaakos:
@fstasi, could you please help with the review?
@per1234, could you please verify the changes once more?

Thank you! 🙏

davegarthsimpson:
confirmed this works for me on mac.

fstasi (changes requested):
@davegarthsimpson seems good but I think you need to update the PR description

fstasi (approved):
lgtm

Issue #1224, opened 2022-07-19
Incorrect editor tab color in the Settings dialog


kittaakos:
### Describe the problem

When the `dark` theme is enabled, the editor tab is still `light` in the `Settings` dialog.

### To reproduce

  - Open the IDE2,
  - If you do not have a `dark` theme selected open the Settings and select a `dark` theme then close with `OK`,
  - Open the Settings,
  - The editor's tab is `light`.

<img width="820" alt="Screen Shot 2022-07-19 at 16 05 00" src="https://user-images.githubusercontent.com/1405703/179770417-1a3ef676-8f01-4e4e-9c63-5e2e9871ebc6.png">
<img width="813" alt="Screen Shot 2022-07-19 at 16 04 46" src="https://user-images.githubusercontent.com/1405703/179770423-645ca53b-5c94-4a2d-884f-71ff0ab63401.png">

### Expected behavior

The editor's tab is `dark` when the theme is `dark`.

### Arduino IDE version

7680194feb4294cca72fd820617113d316d579b3

### Operating system

macOS

### Operating system version

12.3.1

### Issue checklist

- [X] I searched for previous reports in [the issue tracker](https://github.com/arduino/arduino-ide/issues?q=)
- [X] I verified the problem still occurs when using the latest [nightly build](https://github.com/arduino/arduino-ide#nightly-builds)
- [X] My report contains all necessary details

PR #1252, opened 2022-07-25
Custom colors clean up

### LINKED ISSUE #1226: [dev] Clean up the custom color registry items
### Describe the problem

RC-8 could show the board selector with [eight custom colors](https://github.com/arduino/arduino-ide/blob/439cdfbbff382ddc89b652028655e15e381646d8/arduino-ide-extension/src/browser/arduino-frontend-contribution.tsx#L570-L649), the current state fails to do the same with [17 custom color entries](https://github.com/arduino/arduino-ide/blob/2fdb19ea755b729ef384431568509cae3727dd5e/arduino-ide-extension/src/browser/arduino-frontend-contribution.tsx#L184-L351).

Clean it up. Adjust other themes with the fallback.

### To reproduce

This is a code quality issue.

### Expected behavior

The themes work, IDE2 code does not contain unused color entries.

### Arduino IDE version

7680194feb4294cca72fd820617113d316d579b3

### Operating system

macOS

### Operating system version

12.3.1

### Additional context

_No response_

### Issue checklist

francescospissu:
### Motivation
The number of custom colors should be reduced because in the future it will be difficult to support VS Code themes while keeping all these extra colors.
Hard-coded colors must be removed using an appropriate variable.

### Change description
The following is a proposed change taking into account these variables https://code.visualstudio.com/api/references/theme-color.

In particular:

- Remove hard-coded colors using appropriate variables.

- To change `"arduino.output.background"` and `"arduino.output.foreground"` with `"terminal.background"` and `"terminal.foreground"` (https://code.visualstudio.com/api/references/theme-color#integrated-terminal-colors). The values of these variables must be defined.

- `"arduino.branding.primary"` used for the "Learn more" link in the Cloud Sketchbook, could be replaced by `"textLink.foreground"` (https://code.visualstudio.com/api/references/theme-color#text-colors). The values of the new variable must be defined.

- `"arduino.branding.primary"` used for the "Select board and port..." button in the board selector dropdown, could be replaced by `"secondaryButton.foreground"`. This variable is already present and his values reflects those of the older one.

- `"arduino.branding.primary"` used for the check icon in the "Select Board" dialog, can be replaced by  `"list.activeSelectionIconForeground"` (https://code.visualstudio.com/api/references/theme-color#lists-and-trees). The values of the new variable must be defined.

- `"arduino.branding.primary"` used  for the "Select Other Board & Port" string in the "Select Board" dialog, could be replaced by `"editorWidget.foreground"` (https://code.visualstudio.com/api/references/theme-color#editor-widget-colors). The values of the new variable must be defined.

- `"arduino.branding.secondary"` used  for the "NO PORTS DISCOVERED" message in the "Select Board" dialog, could be replaced by `"editorWidget.foreground"`. We can use this variable also for the string "SEARCH BOARD" in the same dialog since "The Editor widget is shown in front of the editor content. Examples are the Find/Replace dialog, the suggestion widget, and the editor hover." (https://code.visualstudio.com/api/references/theme-color#editor-widget-colors).

-  Since the colors of the tabs in the settings dialog are hard-coded, could be possible to use the `tab-activeBackground`, `tab-activeBorder` and `tab-activeForeground` variables (https://code.visualstudio.com/api/references/theme-color#editor-groups-tabs). The values of the new variables must be defined.

- Regarding the board selector dropdown it could be possible to replace: `"arduino.toolbar.dropdown.background"` with `"dropdown.background"`, `"arduino.toolbar.dropdown.label"` with `"dropdown.foreground"`, `"arduino.toolbar.dropdown.border"` with `"dropdown.border"` (https://code.visualstudio.com/api/references/theme-color#dropdown-control). The values are already defined and are corresponding.

- Regarding the active border of the board selector dropdown could be possible to use, as in Theia Blueprint, `"focus.border"` instead of `"arduino.toolbar.dropdown.borderActive"` (https://code.visualstudio.com/api/references/theme-color#base-colors).

- Regarding the board selector drop-down when it's open we can consider the items inside as a list (for the drop-down menus of the Serial Monitor we use the list variables) and we could use these: https://code.visualstudio.com/api/references/theme-color#lists-and-trees, in particular: `"list.hoverBackground"` instead of `"arduino.toolbar.dropdown.option.backgroundHover"`, `"list.activeSelectionForeground"` instead of `"arduino.toolbar.dropdown.option.backgroundSelected"`, `"list.activeSelectionIconForeground"` instead of `"arduino.toolbar.dropdown.iconSelected"`.

Regarding toolbar buttons, it is difficult to find a solution since they are very custom elements. But I noticed that there is also a toolbar in Theia Blueprint:

Theia toolbar

For these items custom colors defined on Theia are used: `"mainToolbar.background"`, `"mainToolbar.foreground"` (https://github.com/eclipse-theia/theia/blob/4b1220ac6602b794ff0369d01acd8b35f910e757/packages/core/src/browser/common-frontend-contribution.ts#L2209-L2223) and `"toolbar.hoverBackground"` defined in https://code.visualstudio.com/api/references/theme-color#action-colors.
It could be possible to use the same variables used for those buttons.

With these changes (without using the variables for the toolbar icons) we reduce the custom colors only to:
`'arduino.toolbar.button.background'`
`'arduino.toolbar.button.hoverBackground'`
`'arduino.toolbar.button.secondary.label'`
`'arduino.toolbar.button.secondary.hoverBackground'`
`'arduino.toolbar.toggleBackground'`

### Other information
Closes #1226 and #1224.

### Reviewer checklist

### DIFF
--- a/arduino-ide-extension/src/browser/arduino-frontend-contribution.tsx
+++ b/arduino-ide-extension/src/browser/arduino-frontend-contribution.tsx
@@ -197,8 +197,8 @@ export class ArduinoFrontendContribution
         id: 'arduino.toolbar.button.hoverBackground',
         defaults: {
           dark: 'button.hoverBackground',
-          light: 'button.foreground',
-          hc: 'textLink.foreground',
+          light: 'button.hoverBackground',
+          hc: 'button.background',
         },
         description:
           'Background color of the toolbar items when hovering over them. Such as Upload, Verify, etc.',
--- a/arduino-ide-extension/src/browser/arduino-frontend-contribution.tsx
+++ b/arduino-ide-extension/src/browser/arduino-frontend-contribution.tsx
@@ -184,24 +184,6 @@ export class ArduinoFrontendContribution
   registerColors(colors: ColorRegistry): void {
     colors.register(
       {
-        id: 'arduino.branding.primary',
-        defaults: {
-          dark: 'statusBar.background',
-          light: 'statusBar.background',
-        },
-        description:
-          'The primary branding color, such as dialog titles, library, and board manager list labels.',
-      },
-      {
-        id: 'arduino.branding.secondary',
-        defaults: {
-          dark: 'statusBar.background',
-          light: 'statusBar.background',
-        },
-        description:
-          'Secondary branding color for list selections, dropdowns, and widget borders.',
-      },
-      {
         id: 'arduino.toolbar.button.background',
         defaults: {
           dark: 'button.background',
@@ -250,92 +232,6 @@ export class ArduinoFrontendContribution
         },
         description:
           'Toggle color of the toolbar items when they are currently toggled (the command is in progress)',
-      },
-      {
-        id: 'arduino.output.foreground',
-        defaults: {
-          dark: 'editor.foreground',
-          light: 'editor.foreground',
-          hc: 'editor.foreground',
-        },
-        description: 'Color of the text in the Output view.',
-      },
-      {
-        id: 'arduino.output.background',
-        defaults: {
-          dark: 'editor.background',
-          light: 'editor.background',
-          hc: 'editor.background',
-        },
-        description: 'Background color of the Output view.',
-      },
-      {
-        id: 'arduino.toolbar.dropdown.border',
-        defaults: {
-          dark: 'dropdown.border',
-          light: 'dropdown.border',
-          hc: 'dropdown.border',
-        },
-        description: 'Border color of the Board Selector.',
-      },
-
-      {
-        id: 'arduino.toolbar.dropdown.borderActive',
-        defaults: {
-          dark: 'focusBorder',
-          light: 'focusBorder',
-          hc: 'focusBorder',
-        },
-        description: "Border color of the Board Selector when it's active",
-      },
-
-      {
-        id: 'arduino.toolbar.dropdown.background',
-        defaults: {
-          dark: 'tab.unfocusedActiveBackground',
-          light: 'tab.unfocusedActiveBackground',
-          hc: 'tab.unfocusedActiveBackground',
-        },
-        description: 'Background color of the Board Selector.',
-      },
-
-      {
-        id: 'arduino.toolbar.dropdown.label',
-        defaults: {
-          dark: 'foreground',
-          light: 'foreground',
-          hc: 'foreground',
-        },
-        description: 'Font color of the Board Selector.',
-      },
-      {
-        id: 'arduino.toolbar.dropdown.iconSelected',
-        defaults: {
-          dark: 'statusBar.background',
-          light: 'statusBar.background',
-          hc: 'statusBar.background',
-        },
-        description:
-          'Color of the selected protocol icon in the Board Selector.',
-      },
-      {
-        id: 'arduino.toolbar.dropdown.option.backgroundHover',
-        defaults: {
-          dark: 'editor.background',
-          light: 'editor.background',
-          hc: 'editor.background',
-        },
-        description: 'Background color on hover of the Board Selector options.',
-      },
-      {
-        id: 'arduino.toolbar.dropdown.option.backgroundSelected',
-        defaults: {
-          dark: 'editor.background',
-          light: 'editor.background',
-          hc: 'editor.background',
-        },
-        description:
-          'Background color of the selected board in the Board Selector.',
       }
     );
   }
--- a/arduino-ide-extension/src/browser/arduino-frontend-contribution.tsx
+++ b/arduino-ide-extension/src/browser/arduino-frontend-contribution.tsx
@@ -232,6 +232,74 @@ export class ArduinoFrontendContribution
         },
         description:
           'Toggle color of the toolbar items when they are currently toggled (the command is in progress)',
+      },
+      {
+        id: 'arduino.toolbar.dropdown.border',
+        defaults: {
+          dark: 'dropdown.border',
+          light: 'dropdown.border',
+          hc: 'dropdown.border',
+        },
+        description: 'Border color of the Board Selector.',
+      },
+
+      {
+        id: 'arduino.toolbar.dropdown.borderActive',
+        defaults: {
+          dark: 'focusBorder',
+          light: 'focusBorder',
+          hc: 'focusBorder',
+        },
+        description: "Border color of the Board Selector when it's active",
+      },
+
+      {
+        id: 'arduino.toolbar.dropdown.background',
+        defaults: {
+          dark: 'tab.unfocusedActiveBackground',
+          light: 'dropdown.background',
+          hc: 'dropdown.background',
+        },
+        description: 'Background color of the Board Selector.',
+      },
+
+      {
+        id: 'arduino.toolbar.dropdown.label',
+        defaults: {
+          dark: 'dropdown.foreground',
+          light: 'dropdown.foreground',
+          hc: 'dropdown.foreground',
+        },
+        description: 'Font color of the Board Selector.',
+      },
+      {
+        id: 'arduino.toolbar.dropdown.iconSelected',
+        defaults: {
+          dark: 'list.activeSelectionIconForeground',
+          light: 'list.activeSelectionIconForeground',
+          hc: 'list.activeSelectionIconForeground',
+        },
+        description:
+          'Color of the selected protocol icon in the Board Selector.',
+      },
+      {
+        id: 'arduino.toolbar.dropdown.option.backgroundHover',
+        defaults: {
+          dark: 'list.hoverBackground',
+          light: 'list.hoverBackground',
+          hc: 'list.hoverBackground',
+        },
+        description: 'Background color on hover of the Board Selector options.',
+      },
+      {
+        id: 'arduino.toolbar.dropdown.option.backgroundSelected',
+        defaults: {
+          dark: 'list.activeSelectionBackground',
+          light: 'list.activeSelectionBackground',
+          hc: 'list.activeSelectionBackground',
+        },
+        description:
+          'Background color of the selected board in the Board Selector.',
       }
     );
   }
--- a/arduino-ide-extension/src/browser/arduino-frontend-contribution.tsx
+++ b/arduino-ide-extension/src/browser/arduino-frontend-contribution.tsx
@@ -202,16 +202,6 @@ export class ArduinoFrontendContribution
           'Secondary branding color for list selections, dropdowns, and widget borders.',
       },
       {
-        id: 'arduino.foreground',
-        defaults: {
-          dark: 'editorWidget.background',
-          light: 'editorWidget.background',
-          hc: 'editorWidget.background',
-        },
-        description:
-          'Color of the Arduino IDE foreground which is used for dialogs, such as the Select Board dialog.',
-      },
-      {
         id: 'arduino.toolbar.button.background',
         defaults: {
           dark: 'button.background',
--- a/arduino-ide-extension/src/browser/data/dark.color-theme.json
+++ b/arduino-ide-extension/src/browser/data/dark.color-theme.json
@@ -8,6 +8,7 @@
     "list.inactiveSelectionForeground": "#dae3e3",
     "list.inactiveSelectionBackground": "#434f54",
     "list.hoverBackground": "#1f272a",
+    "list.activeSelectionIconForeground": "#0ca1a6",
     "progressBar.background": "#005c5f",
     "editor.background": "#1f272a",
     "editor.foreground": "#dae3e3",
@@ -16,6 +17,7 @@
     "editorCursor.foreground": "#434f54",
     "editorWhitespace.foreground": "#bfbfbf",
     "editorWidget.background": "#171e21",
+    "editorWidget.foreground": "#dae3e3",
     "focusBorder": "#dae3e3",
     "menubar.selectionBackground": "#ffffff",
     "menubar.selectionForeground": "#212121",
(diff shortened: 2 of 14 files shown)


fstasi (approved):
Ok, so there's a lot to analyze here...

> Remove hard-coded colors using appropriate variables.

Seems we should and I don't see why we shouldn't. ✅ 

> To change "arduino.output.background" and "arduino.output.foreground" with "terminal.background" and "terminal.foreground" (https://code.visualstudio.com/api/references/theme-color#integrated-terminal-colors). The values of these variables must be defined.

I don't think there is a reason why they should be different. At least not as for now. @91volt do you think we are going to have different colours for terminal and output in the immediate future? If not, I'd ✅ 

> "arduino.branding.primary" used for the "Learn more" link in the Cloud Sketchbook, could be replaced by "textLink.foreground" (https://code.visualstudio.com/api/references/theme-color#text-colors). The values of the new variable must be defined.

I don't see specific pitfalls here. If we are planning to do crazy stuff with `textLink` we can just set `arduino.branding.primary` default to `textLink.foreground`. Unless really needed, let's complicate the things when it's needed. Not before. ✅ here as well for me 

> "arduino.branding.primary" used for the "Select board and port..." button in the board selector dropdown, could be replaced by "secondaryButton.foreground". This variable is already present and his values reflects those of the older one.

The `arduino.branding.primary` has no contextual meaning, hence `secondaryButton.foreground` it's probably better as it provides some semantic information. ✅ as long as it does not clash with other colors, but reading your comment I think it does not.

> "arduino.branding.primary" used for the check icon in the "Select Board" dialog, can be replaced by "list.activeSelectionIconForeground" (https://code.visualstudio.com/api/references/theme-color#lists-and-trees). The values of the new variable must be defined.

I'm a little skeptical here, because we are using a different, yet similar, semantic name. It's also true that `arduino.branding.primary` has no meaning at all. So can we decide between these two options ❓ 
- use `list.activeSelectionIconForeground` (but please comment in the code that using this color can be considered "improper" 
- add a new semantic "arduino" variable that defaults to `list.activeSelectionIconForeground` when it's not set in a theme

> "arduino.branding.primary" used for the "Select Other Board & Port" string in the "Select Board" dialog, could be replaced by "editorWidget.foreground" (https://code.visualstudio.com/api/references/theme-color#editor-widget-colors). The values of the new variable must be defined.

This seems good to me ✅ 

> "arduino.branding.secondary" used for the "NO PORTS DISCOVERED" message in the "Select Board" dialog, could be replaced by "editorWidget.foreground". We can use this variable also for the string "SEARCH BOARD" in the same dialog since "The Editor widget is shown in front of the editor content. Examples are the Find/Replace dialog, the suggestion widget, and the editor hover." (https://code.visualstudio.com/api/references/theme-color#editor-widget-colors).

Not sure I get this one. will the change you suggest have impacts on the current colors ❓ 

> Since the colors of the tabs in the settings dialog are hard-coded, could be possible to use the tab-activeBackground, tab-activeBorder and tab-activeForeground variables (https://code.visualstudio.com/api/references/theme-color#editor-groups-tabs). The values of the new variables must be defined.

This seems quite straightforward to me, I'd say yes ✅ 

> Regarding the board selector dropdown it could be possible to replace: "arduino.toolbar.dropdown.background" with "dropdown.background", "arduino.toolbar.dropdown.label" with "dropdown.foreground", "arduino.toolbar.dropdown.border" with "dropdown.border" (https://code.visualstudio.com/api/references/theme-color#dropdown-control). The values are already defined and are corresponding.

If they are already defined and corresponding, I'd say yes ✅ . In the future we might need to differentiate between the two, but let's keep the codebase simple as long as we can.

> Regarding the active border of the board selector dropdown could be possible to use, as in Theia Blueprint, "focus.border" instead of "arduino.toolbar.dropdown.borderActive" (https://code.visualstudio.com/api/references/theme-color#base-colors).

I think it's ok, but does it mean changing the `focus.border` color in order to achieve the same result. If so, we need to double check it's not used (or if used does not break stuff). So questionmark ❓  here

> Regarding the board selector drop-down when it's open we can consider the items inside as a list (for the drop-down menus of the Serial Monitor we use the list variables) and we could use these: https://code.visualstudio.com/api/references/theme-color#lists-and-trees, in particular: "list.hoverBackground" instead of "arduino.toolbar.dropdown.option.backgroundHover", "list.activeSelectionForeground" instead of "arduino.toolbar.dropdown.option.backgroundSelected", "list.activeSelectionIconForeground" instead of "arduino.toolbar.dropdown.iconSelected".

Similar comment as for the "Select Board" . You are proposing to use a similar, yet slightly different semantic meaning. we can probably set those you suggest as defaults, just for the sake of clarity. what do you think ❓

fstasi (commented):
great work by the way. I definitely see the value and you inspected all the options thoroughly .

francescospissu:
Thanks for you review @fstasi! Below I try to explain all the considerations I have made for the points that are not clear.

> > "arduino.branding.primary" used for the check icon in the "Select Board" dialog, can be replaced by "list.activeSelectionIconForeground" (https://code.visualstudio.com/api/references/theme-color#lists-and-trees). The values of the new variable must be defined.
> 
> I'm a little skeptical here, because we are using a different, yet similar, semantic name. It's also true that arduino.branding.primary has no meaning at all. So can we decide between these two options ❓
> 
> - use list.activeSelectionIconForeground (but please comment in the code that using this color can be considered "improper"
> - add a new semantic "arduino" variable that defaults to list.activeSelectionIconForeground when it's not set in a theme

`"list.activeSelectionIconForeground"` variable has been chosen since it identifies the foreground of an icon that belongs to an active element of a list:

Screenshot 2022-07-26 164210

> > "arduino.branding.secondary" used for the "NO PORTS DISCOVERED" message in the "Select Board" dialog, could be replaced by "editorWidget.foreground". We can use this variable also for the string "SEARCH BOARD" in the same dialog since "The Editor widget is shown in front of the editor content. Examples are the Find/Replace dialog, the suggestion widget, and the editor hover." (https://code.visualstudio.com/api/references/theme-color#editor-widget-colors).
> 
> Not sure I get this one. will the change you suggest have impacts on the current colors ❓


The variable `"editorWidget.foreground"` has been chosen taking into account the design specifications on Figma (actually the current implementation is not aligned with that) where the strings "Boards", "Ports" and "NO PORT" have the  same color.

So, yes, this change has an impact on current colors. The current state is the one shown in the attached image for the previous point.

This is the suggested change:

Screenshot 2022-07-26 164210

and this is the requirement:
Screenshot 2022-07-28 105015

> > Regarding the active border of the board selector dropdown could be possible to use, as in Theia Blueprint, "focus.border" instead of "arduino.toolbar.dropdown.borderActive" (https://code.visualstudio.com/api/references/theme-color#base-colors).
> 
> I think it's ok, but does it mean changing the focus.border color in order to achieve the same result. If so, we need to double check it's not used (or if used does not break stuff). So questionmark ❓ here

A custom value for `"focusBorder"` is already defined (default: `"focusBorder": "#7fcbcd"`, dark: `"focusBorder": "#dae3e3"`) and it is used for the Serial Monitor drop-down:

Screenshot 2022-07-28 105015

> > Regarding the board selector drop-down when it's open we can consider the items inside as a list (for the drop-down menus of the Serial Monitor we use the list variables) and we could use these: https://code.visualstudio.com/api/references/theme-color#lists-and-trees, in particular: "list.hoverBackground" instead of "arduino.toolbar.dropdown.option.backgroundHover", "list.activeSelectionForeground" instead of "arduino.toolbar.dropdown.option.backgroundSelected", "list.activeSelectionIconForeground" instead of "arduino.toolbar.dropdown.iconSelected".
> 
> Similar comment as for the "Select Board" . You are proposing to use a similar, yet slightly different semantic meaning. we can probably set those you suggest as defaults, just for the sake of clarity. what do you think ❓

This is the trickiest part since I assumed to consider the open board selector drop-down menu as a list (like the Serial Monitor drop-down menu in the image from the previous point). For this point I'd wait for the opinion of @91volt.

91volt:
> Remove hard-coded colors using appropriate variables.

✅

> "arduino.branding.primary" used for the "Learn more" link in the Cloud Sketchbook, could be replaced by "textLink.foreground" (https://code.visualstudio.com/api/references/theme-color#text-colors). The values of the new variable must be defined.

✅


> To change "arduino.output.background" and "arduino.output.foreground" with "terminal.background" and "terminal.foreground" (https://code.visualstudio.com/api/references/theme-color#integrated-terminal-colors). The values of these variables must be defined.

✅


> "arduino.branding.primary" used for the check icon in the "Select Board" dialog, can be replaced by "list.activeSelectionIconForeground" (https://code.visualstudio.com/api/references/theme-color#lists-and-trees). The values of the new variable must be defined.

✅  Works for me **TASK FOR ME**


> "arduino.branding.primary" used for the "Select Other Board & Port" string in the "Select Board" dialog, could be replaced by "editorWidget.foreground" (https://code.visualstudio.com/api/references/theme-color#editor-widget-colors). The values of the new variable must be defined.

🟡 -> I believe the identified token might be misleading, and we should address this using just one dialog title in dark gray as all other dialogs, it's redundant to keep both titles. 


> "arduino.branding.secondary" used for the "NO PORTS DISCOVERED" message in the "Select Board" dialog, could be replaced by "editorWidget.foreground". We can use this variable also for the string "SEARCH BOARD" in the same dialog since "The Editor widget is shown in front of the editor content. Examples are the Find/Replace dialog, the suggestion widget, and the editor hover." (https://code.visualstudio.com/api/references/theme-color#editor-widget-colors).

🔴 -> I think that our dialog shouldn't be confused with internal editor's widgets such as find and replace. We should have a contextual variable for arduino dialogs subtle foregrounds. Token: arduino.dialog.foreground.subtle. 


> Since the colors of the tabs in the settings dialog are hard-coded, could be possible to use the tab-activeBackground, tab-activeBorder and tab-activeForeground variables (https://code.visualstudio.com/api/references/theme-color#editor-groups-tabs). The values of the new variables must be defined.

✅  If they are not in usage works for me, and in case ->**TASK FOR ME**


> Regarding the board selector dropdown it could be possible to replace: "arduino.toolbar.dropdown.background" with "dropdown.background", "arduino.toolbar.dropdown.label" with "dropdown.foreground", "arduino.toolbar.dropdown.border" with "dropdown.border" (https://code.visualstudio.com/api/references/theme-color#dropdown-control). The values are already defined and are corresponding.
Regarding the active border of the board selector dropdown could be possible to use, as in Theia Blueprint, "focus.border" instead of "arduino.toolbar.dropdown.borderActive" (https://code.visualstudio.com/api/references/theme-color#base-colors).
Regarding the board selector drop-down when it's open we can consider the items inside as a list (for the drop-down menus of the Serial Monitor we use the list variables) and we could use these: https://code.visualstudio.com/api/references/theme-color#lists-and-trees, in particular: "list.hoverBackground" instead of "arduino.toolbar.dropdown.option.backgroundHover", "list.activeSelectionForeground" instead of "arduino.toolbar.dropdown.option.backgroundSelected", "list.activeSelectionIconForeground" instead of "arduino.toolbar.dropdown.iconSelected".
For these items custom colors defined on Theia are used: "mainToolbar.background", "mainToolbar.foreground" (https://github.com/eclipse-theia/theia/blob/4b1220ac6602b794ff0369d01acd8b35f910e757/packages/core/src/browser/common-frontend-contribution.ts#L2209-L2223) and "toolbar.hoverBackground" defined in https://code.visualstudio.com/api/references/theme-color#action-colors.

🔴 Since the arduino toolbar and board selector are our own custom components I would keep them autonomous from other vscode tokens. -> But all these tokens you found are useful fallbacks for compatibility with other themes



> It could be possible to use the same variables used for those buttons.
With these changes (without using the variables for the toolbar icons) we reduce the custom colors only to:
'arduino.toolbar.button.background'
'arduino.toolbar.button.hoverBackground'
'arduino.toolbar.button.secondary.label'
'arduino.toolbar.button.secondary.hoverBackground'
 'arduino.toolbar.toggleBackground'

🔴 -> Same as toolbar and dropdown I believe we should keep them autonomous and I would like to keep the granularity for each button, primary and secondary. -> Same as before, this are useful fallbacks.


This is my point of view that takes into account the usage of the color tokens shared with the web editor in the future.
In any case great work Francesco, I believe is a great cleanup overall.

91volt:
> "editorWidget.foreground" was chosen because I considered our dialogs as widgets shown in front of the editor content and also because it is not already used. If we want to have a contextual variable for this purpose maybe two other variables are also needed to replace "editorWidget.background" and "editorWidget.border", in this case we are increasing the number of custom colors by adding: "arduino.dialog.foreground.subtle", "arduino.dialog.background" (or equivalent) and "arduino.dialog.border" (or equivalent) .

✅ Ok, thanks for highlighting that we are already using editorWidget tokens for dialogs. 
So I change my mind in this case and I agree with your change

> Do you think we can use these for both purposes @91volt?
✅ Yes it looks feasible since in our dark and light themes the IDE main area background color is the same - or very close - to the dialog bg. At a certain point we should replace the preferences panel, so we won't have anymore this kind of tabs navigation in the dialog, and this sounds the best way to avoid adding variables.

> About this, my assumption was to consider the board selector drop-down menu as a "standard" drop-down menu which, when expanded, presents the items inside as a list. This is probably not correct from a design point of view, so it's possible to revert to the previous behavior using the custom variables:

Yep I confirm would like to revert to the previous state of tokens here.


> This is also forced from my point of view in fact I didn't applied these changes.

👌 Perfect

francescospissu:
Merging this PR the number of custom colors has been reduced to 12 items:

`"arduino.toolbar.dropdown.background"`
`"arduino.toolbar.dropdown.label" `
`"arduino.toolbar.dropdown.border"`
`"arduino.toolbar.dropdown.borderActive"`
`"arduino.toolbar.dropdown.option.backgroundHover"`
`"arduino.toolbar.dropdown.option.backgroundSelected"`
`"arduino.toolbar.dropdown.iconSelected"`
`"arduino.toolbar.button.background"`
`"arduino.toolbar.button.hoverBackground"`
`"arduino.toolbar.button.secondary.label"`
`"arduino.toolbar.button.secondary.hoverBackground"`
`"arduino.toolbar.toggleBackground"`

Issue #1205, opened 2022-07-18
The min window size is fixed in IDE2, but the Settings dialog does not scale well


kittaakos:
### Describe the problem

When the IDE2 window size uses the minimum height and width, and the UI scale is > 100%, then the _Settings_ dialog looks broken.

### To reproduce

 - Start the IDE2,
 - Increase the UI scale (`Cmd` + `+` on macOS),
 - Open the _Settings_ dialog.
 - The buttons are off from the dialog.

https://user-images.githubusercontent.com/1405703/179472660-bdd12eba-adc6-4eae-a1c7-21b708be42f7.mp4

### Expected behavior

Settings dialog can scale.

### Arduino IDE version

ed41b258

### Operating system

macOS

### Operating system version

12.3.1

### Issue checklist

- [X] I searched for previous reports in [the issue tracker](https://github.com/arduino/arduino-ide/issues?q=)
- [X] I verified the problem still occurs when using the latest [nightly build](https://github.com/arduino/arduino-ide#nightly-builds)
- [X] My report contains all necessary details

PR #1256, opened 2022-07-26
#1205: improve "broken UI" with settings dialog when app is scaled up

davegarthsimpson:
### Motivation
Settings dialog reported as visually unpleasing, and functionally impaired when scaling up the IDE.

### Change description
Uses a fixed footer and `overflow: auto` for settings dialog content.

### Reviewer checklist

### DIFF
--- a/arduino-ide-extension/src/browser/style/settings-dialog.css
+++ b/arduino-ide-extension/src/browser/style/settings-dialog.css
@@ -6,6 +6,11 @@
 
 #arduino-settings-dialog-container > .dialogBlock > .dialogContent {
     justify-content: flex-start;
+    overflow: auto;
+}
+
+#arduino-settings-dialog-container > .dialogBlock > .dialogControl {
+    padding: 16px 0 26px;
 }
 
 .arduino-settings-dialog .content {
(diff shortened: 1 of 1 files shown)


kittaakos:
I've tried it, and I think this is a good solution in IDE2 👍

francescospissu (approved):
It works as expected 👍

AlbyIanna:
The issue is spread across all the dialogs.

This fixes #1205, but doesn't solve the issue for other dialogs.

e.g.: the Select Board dialog
image


I believe we should solve it at the root. [Here](https://github.com/arduino/arduino-ide/compare/dialog-ui-scale) I made an experiment, but it raised some other scrolling issues, so I didn't open a PR.

davegarthsimpson:
> The issue is spread across all the dialogs.
> 
> This PR fixes #1205, but doesn't solve the issue for other dialogs.
> 
> e.g.: the Select Board dialog image
> 
> I believe we should solve it at the root. [Here](https://github.com/arduino/arduino-ide/compare/dialog-ui-scale) I made an experiment, but it raised some other scrolling issues, so I didn't open a PR.

How many dialogs do we have? If only a handful I would solve case by case, as we may want a different design / UX for each, e.g. it's clearly not nice if scrollable elements also have a scrollable parent.

fstasi (approved):

PR #1261, opened 2022-07-27
Fix current IDE language on start-up

### LINKED ISSUE #1219: IDE2 might forget the selected language
### Describe the problem

From time to time, IDE2 forgets the user-selected language:
 - Open IDE2 select `de` as the default,
 - IDE2 reloads, it shows `de`,
 - Quit IDE2,
 - Start IDE2, it still shows `de` ✅ ,
 - Select `zh-cn` as the default language,
 - IDE2 reloads and shows `zh-cn` ✅ ,
 - Quit IDE2,
 - Open IDE2,
 - The language is again `en` 🐛 

### To reproduce

See the description for the steps.

https://user-images.githubusercontent.com/1405703/179749657-e8988cdb-3613-4042-a095-d90d9a233383.mp4

### Expected behavior

IDE2 remembers the language.

### Arduino IDE version

1f7c2eb5

### Operating system

macOS

### Operating system version

12.3.1

### Additional context

_No response_

### Issue checklist

AlbyIanna:
### Motivation
[IDE2 might forget the selected language](https://github.com/arduino/arduino-ide/issues/1219)

I noticed the root of the problem is here:

https://github.com/eclipse-theia/theia/blob/f4327c2f0c0f47f960a302590a51698879357841/packages/core/src/node/i18n/localization-backend-contribution.ts#L38-L39

`req.params.locale` is always restored correctly, but `this.localizationProvider.getAvailableLanguages()` doesn't return all the available languages. This happens because `getAvailableLanguages()` returns all the registered localizations with `languagePack` set to `true` (unless we call `getAvailableLanguages(true)`, in that case, it would always return all the languages):

https://github.com/eclipse-theia/theia/blob/f4327c2f0c0f47f960a302590a51698879357841/packages/core/src/node/i18n/localization-provider.ts#L50

The weird thing is that this happens also with _some_ languages registered with a language pack (so they'll eventually have `languagePack` set to `true`). I say _some_ languages because _some others_ work as expected. I strongly believe this is happening because of a race condition happening because `getAvailableLanguages` is called before the language packs are finished loading.

EDIT:
The root cause of this appears to be that `PluginDeployer` completes the execution of `deployPlugins` after the express backend tries to restore the current language

https://github.com/eclipse-theia/theia/blob/26649452aeac342f7551b8cae502740799085772/packages/plugin-ext/src/main/node/plugin-deployer-impl.ts#L269

The solution I've found is to wait for [this event to emit](https://github.com/eclipse-theia/theia/blob/26649452aeac342f7551b8cae502740799085772/packages/plugin-ext/src/main/node/plugin-deployer-impl.ts#L291)
 
### Change description
Re-implemented the restoration of the language in `localization-backend-contribution.ts`

Wait for the plugins to complete the deploy before restoring the current language:
https://github.com/arduino/arduino-ide/pull/1261/files#diff-19d891f1ffb8ce2979b6bee0d166700e20a6b965d1be5f72b3e331d0442eba35R16-R18

### Other information
This PR is a workaround for a Theia issue.
I opened an issue in the Theia repo here: https://github.com/eclipse-theia/theia/issues/11471
As soon as Theia fixes it, we can remove this code and the rebinding.

I've also opened a PR on Theia here: https://github.com/eclipse-theia/theia/pull/11472

Fixes #1219

### Reviewer checklist

### DIFF
--- a/arduino-ide-extension/src/node/i18n/localization-backend-contribution.ts
+++ b/arduino-ide-extension/src/node/i18n/localization-backend-contribution.ts
@@ -1,0 +1,45 @@
+import * as express from 'express';
+import { inject, injectable } from '@theia/core/shared/inversify';
+import { LocalizationBackendContribution as TheiaLocalizationBackendContribution } from '@theia/core/lib/node/i18n/localization-backend-contribution';
+import { PluginDeployer } from '@theia/plugin-ext/lib/common/plugin-protocol';
+import { PluginDeployerImpl } from '@theia/plugin-ext/lib/main/node/plugin-deployer-impl';
+import { Deferred } from '@theia/core/lib/common/promise-util';
+
+@injectable()
+export class LocalizationBackendContribution extends TheiaLocalizationBackendContribution {
+  @inject(PluginDeployer)
+  private readonly pluginDeployer: PluginDeployerImpl;
+
+  private readonly initialized = new Deferred<void>();
+
+  override async initialize(): Promise<void> {
+    this.pluginDeployer.onDidDeploy(() => {
+      this.initialized.resolve();
+    });
+    return super.initialize();
+  }
+
+  override configure(app: express.Application): void {
+    app.get('/i18n/:locale', async (req, res) => {
+      let locale = req.params.locale;
+      /*
+        Waiting for the deploy of the language plugins is neecessary to avoid checking the available
+        languages before they're finished to be loaded: https://github.com/eclipse-theia/theia/issues/11471 
+      */
+      const start = performance.now();
+      await this.initialized.promise;
+      console.info(
+        'Waiting for the deploy of the language plugins took: ' +
+          (performance.now() - start) +
+          ' ms.'
+      );
+      locale = this.localizationProvider
+        .getAvailableLanguages()
+        .some((e) => e.languageId === locale)
+        ? locale
+        : 'en';
+      this.localizationProvider.setCurrentLanguage(locale);
+      res.send(this.localizationProvider.loadLocalization(locale));
+    });
+  }
+}
--- a/arduino-ide-extension/src/node/arduino-ide-backend-module.ts
+++ b/arduino-ide-extension/src/node/arduino-ide-backend-module.ts
@@ -82,7 +82,7 @@ import {
 } from '../common/protocol/authentication-service';
 import { ArduinoFirmwareUploaderImpl } from './arduino-firmware-uploader-impl';
 import { PlotterBackendContribution } from './plotter/plotter-backend-contribution';
-import { ArduinoLocalizationContribution } from './arduino-localization-contribution';
+import { ArduinoLocalizationContribution } from './i18n/arduino-localization-contribution';
 import { LocalizationContribution } from '@theia/core/lib/node/i18n/localization-contribution';
 import { MonitorManagerProxyImpl } from './monitor-manager-proxy-impl';
 import { MonitorManager, MonitorManagerName } from './monitor-manager';
@@ -102,6 +102,8 @@ import WebSocketProviderImpl from './web-socket/web-socket-provider-impl';
 import { WebSocketProvider } from './web-socket/web-socket-provider';
 import { ClangFormatter } from './clang-formatter';
 import { FormatterPath } from '../common/protocol/formatter';
+import { LocalizationBackendContribution } from './i18n/localization-backend-contribution';
+import { LocalizationBackendContribution as TheiaLocalizationBackendContribution } from '@theia/core/lib/node/i18n/localization-backend-contribution';
 
 export default new ContainerModule((bind, unbind, isBound, rebind) => {
   bind(BackendApplication).toSelf().inSingletonScope();
@@ -395,4 +397,8 @@ export default new ContainerModule((bind, unbind, isBound, rebind) => {
   bind(BackendApplicationContribution).toService(PlotterBackendContribution);
   bind(ArduinoLocalizationContribution).toSelf().inSingletonScope();
   bind(LocalizationContribution).toService(ArduinoLocalizationContribution);
+  bind(LocalizationBackendContribution).toSelf().inSingletonScope();
+  rebind(TheiaLocalizationBackendContribution).toService(
+    LocalizationBackendContribution
+  );
 });
--- a/arduino-ide-extension/src/node/i18n/arduino-localization-contribution.ts
+++ b/arduino-ide-extension/src/node/i18n/arduino-localization-contribution.ts
@@ -11,142 +11,142 @@ export class ArduinoLocalizationContribution
   async registerLocalizations(registry: LocalizationRegistry): Promise<void> {
     registry.registerLocalizationFromRequire(
       'af',
-      require('../../build/i18n/af.json')
+      require('../../../build/i18n/af.json')
     );
 
     registry.registerLocalizationFromRequire(
       'en',
-      require('../../build/i18n/en.json')
+      require('../../../build/i18n/en.json')
     );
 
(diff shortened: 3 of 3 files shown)


kittaakos (changes requested):
It does not work for me. I tried the steps from #1219.

https://user-images.githubusercontent.com/1405703/181538770-aa72fe3d-9e67-4cb3-bf7d-9a70679d25e5.mp4

AlbyIanna:
Thanks for testing it @kittaakos. I've just updated a PR with a new fix, now it should work.

arduino-ide-extension/src/node/i18n/localization-backend-contribution.ts
```diff
@@ -0,0 +1,39 @@
...
+import { inject, injectable } from '@theia/core/shared/inversify';
+import { LocalizationBackendContribution as TheiaLocalizationBackendContribution } from '@theia/core/lib/node/i18n/localization-backend-contribution';
+import { PluginDeployer } from '@theia/plugin-ext/lib/common/plugin-protocol';
+import { PluginDeployerImpl } from '@theia/plugin-ext/lib/main/node/plugin-deployer-impl';
+import { Deferred } from '@theia/core/lib/common/promise-util';
+
+@injectable()
+export class LocalizationBackendContribution extends TheiaLocalizationBackendContribution {
+  @inject(PluginDeployer)
+  protected readonly pluginDeployer: PluginDeployerImpl;
```
kittaakos, line 11: can be `private`.

arduino-ide-extension/src/node/i18n/localization-backend-contribution.ts
```diff
@@ -0,0 +1,39 @@
...
+
+  private initialized = new Deferred<void>();
```
kittaakos, line 13: should be `readonly`

arduino-ide-extension/src/node/i18n/localization-backend-contribution.ts
```diff
@@ -0,0 +1,39 @@
...
+
+  override async initialize(): Promise<void> {
+    this.pluginDeployer.onDidDeploy(() => {
+      this.initialized.resolve();
+    });
+    await this.localizationRegistry.initialize();
```
kittaakos, line 19: Why not calling `super.initialize()`. The default Theia behavior is what IDE2 wants to use besides the listener.

arduino-ide-extension/src/node/i18n/localization-backend-contribution.ts
```diff
@@ -0,0 +1,39 @@
...
+
+@injectable()
+export class LocalizationBackendContribution extends TheiaLocalizationBackendContribution {
+  @inject(PluginDeployer)
+  protected readonly pluginDeployer: PluginDeployerImpl;
+
+  private initialized = new Deferred<void>();
+
+  override async initialize(): Promise<void> {
+    this.pluginDeployer.onDidDeploy(() => {
```
kittaakos, line 16: It's an OK solution for now, but the deferred promise will never resolve if the plugins are deployed by the time this `initialize` function is called.

arduino-ide-extension/src/node/i18n/localization-backend-contribution.ts
```diff
@@ -0,0 +1,39 @@
...
+  }
+
+  override configure(app: express.Application): void {
+    app.get('/i18n/:locale', async (req, res) => {
+      let locale = req.params.locale;
+      /*
+        Waiting for the deploy of the language plugins is neecessary to avoid checking the available
+        languages before they're finished to be loaded: https://github.com/eclipse-theia/theia/issues/11471 
+      */
+      await this.initialized.promise;
```
kittaakos, line 29: Just wondering, did you check if it slows down the IDE2 startup? The frontend must wait for this, right?

kittaakos (commented):
Besides a few remarks, it's looking great. Please react to them, and I will do the review afterward with the final changeset. Thank you!

arduino-ide-extension/src/node/i18n/localization-backend-contribution.ts
davegarthsimpson: @kittaakos is that a likely scenario? (getting to this initialize post deploy)

kittaakos: No, I do not think so.

Based on the method names (`initialize` and `start`), I do not think it will cause an issue at runtime, but I did not thoroughly check the Theia code.

davegarthsimpson: @AlbyIanna lets keep note of this, maybe we'll get a hint [here](https://github.com/eclipse-theia/theia/pull/11472) anyway.

arduino-ide-extension/src/node/i18n/localization-backend-contribution.ts
AlbyIanna: I'm not sure how to calculate this.

With the current build, I can see these two log lines:
```
root INFO [2c0ce3a6-ca5a-4f40-b831-533e06823a9f] Load contributions of 16 plugins: 86.5 ms [Finished 4.873 s after frontend start]
root INFO [2c0ce3a6-ca5a-4f40-b831-533e06823a9f] Start of 16 plugins: 575.4 ms [Finished 5.460 s after frontend start]
```

Then, in this case, I think the frontend is slowed down by `86.5 ms + 574.4 ms = 660.9 ms`

Am I correct?

kittaakos:
```ts
const start = performance.now();
await this.initialized.promise;
console.log('took: ' + (performance.now() - start) + ' ms')
```

AlbyIanna:
```
root INFO took: 405.16497199982405 ms
```

Quite significant. I suppose this delay will increase as we keep adding new language packs.

davegarthsimpson: > This should be done fairly quickly without impacting the load time on Electron
from: [here](https://github.com/eclipse-theia/theia/pull/11472#pullrequestreview-1053143796)

davegarthsimpson: maybe it's worth measuring the impact in the build (if the above is from source)? (maybe there are some optimisations that might speed up the plugin deployment)

arduino-ide-extension/src/node/i18n/localization-backend-contribution.ts
```diff
@@ -0,0 +1,39 @@
...
+  @inject(PluginDeployer)
+  private readonly pluginDeployer: PluginDeployerImpl;
+
+  private readonly initialized = new Deferred<void>();
+
+  override async initialize(): Promise<void> {
+    this.pluginDeployer.onDidDeploy(() => {
+      this.initialized.resolve();
+    });
+    super.initialize();
```
kittaakos, line 19: This is a bug. You either must return with `super.initialize() or `await` for it.

arduino-ide-extension/src/node/i18n/localization-backend-contribution.ts
AlbyIanna: I've pushed the code to log the performance so that we see how the time grows as we keep adding new language packs.

kittaakos (approved):
This is working great ✅ 

Thank you so much for tracking down the bug upstream, opening an issue, and proposing fix for the Theia community. Excellent effort 💪

davegarthsimpson (approved):
Nice one @AlbyIanna

PR #1264, opened 2022-07-28
remove state from stepper input and simplify

davegarthsimpson:
### Motivation
The `settings-step-input.tsx` component contains unnecessarily verbose code and inefficient logic.

### Change description
Refactors the code, to exclude useState and useEffect hooks.

### Acceptance Criteria
The stepper component should function exactly as before (prior to the changes in this PR).

### Reviewer checklist

### DIFF
--- a/arduino-ide-extension/src/browser/dialogs/settings/settings-step-input.tsx
+++ b/arduino-ide-extension/src/browser/dialogs/settings/settings-step-input.tsx
@@ -1,5 +1,6 @@
 import * as React from '@theia/core/shared/react';
 import classnames from 'classnames';
+import _ = require('lodash');
 
 interface SettingsStepInputProps {
   value: number;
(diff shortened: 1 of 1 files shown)


arduino-ide-extension/src/browser/dialogs/settings/settings-step-input.tsx
```diff
@@ -1,5 +1,6 @@
 import * as React from '@theia/core/shared/react';
 import classnames from 'classnames';
+import _ = require('lodash');
```
kittaakos, line 3: IDE2 does not directly depend on `lodash`, so this code can technically break.

```
yarn why lodash
yarn why v1.22.18
[1/4] 🤔  Why do we have the module "lodash"...?
[2/4] 🚚  Initialising dependency graph...
warning Resolution field "@types/react@16.14.25" is incompatible with requested version "@types/react@^17"
[3/4] 🔍  Finding dependency...
[4/4] 🚡  Calculating file sizes...
=> Found "lodash@4.17.21"
info Reasons this module exists
   - "_project_#style-dictionary" depends on it
   - Hoisted from "_project_#style-dictionary#lodash"
   - Hoisted from "_project_#arduino-ide-extension#typemoq#lodash"
   - Hoisted from "_project_#electron-app#@theia#debug#requestretry#lodash"
   - Hoisted from "_project_#@theia#cli#mocha#yargs-unparser#lodash"
   - Hoisted from "_project_#jsdom#request-promise-native#request-promise-core#lodash"
   - Hoisted from "_project_#@theia#cli#@theia#ffmpeg#@electron#get#global-tunnel-ng#lodash"
   - Hoisted from "_project_#lerna#@lerna#clean#@lerna#prompt#inquirer#lodash"
   - Hoisted from "_project_#lerna#@lerna#version#@lerna#conventional-commits#conventional-changelog-core#lodash"
   - Hoisted from "_project_#lerna#@lerna#version#@lerna#conventional-commits#conventional-changelog-core#conventional-changelog-writer#lodash"
   - Hoisted from "_project_#lerna#@lerna#version#@lerna#conventional-commits#conventional-changelog-core#conventional-commits-parser#lodash"
info Disk size without dependencies: "4.88MB"
info Disk size with unique dependencies: "4.88MB"
info Disk size with transitive dependencies: "4.88MB"
info Number of shared dependencies: 0
✨  Done in 1.09s.
```

Could you please check that importing the entire `lodash` increased the frontend size by almost 5MB or was it already required?

```
info Disk size without dependencies: "4.88MB"
```

kittaakos: > or was it already required?

It was.

with lodash 11.767.954 bytes (11,8 MB on disk)
without lodash 11.698.043 bytes (11,7 MB on disk)

davegarthsimpson: lodash is not a deal breaker here, the motivation for the PR is the reacty stuff, I thought it was a direct dependency...thanks for catching this

arduino-ide-extension/src/browser/dialogs/settings/settings-step-input.tsx
```diff
@@ -86,34 +52,14 @@ const SettingsStepInput: React.FC<SettingsStepInputProps> = (
     const number = Number(eventValue);
 
     if (!isNaN(number) && number !== value) {
-      let newValue;
```
kittaakos, line 89: I assume the code could return with the function above 👆

```ts
if (eventValue === '') {
  setSettingsStateValue(0);
  return; // <---  this?
}
```

kittaakos (approved):
I tried it and it worked. Thank you!

AlbyIanna:
The code looks fine to me overall, but I noticed a little bug.

It is possible, via shortcuts (CMD/CTRL + '+', CMD/CTRL + '-'), to exceed the min/max thresholds. For example, if the interface scale is set to `200` I can press CMD + '+' to increase the scale to `220`. When the threshold is exceeded, the stepper buttons might look enabled even if they should be disabled.

https://user-images.githubusercontent.com/6939054/181765608-94f5dba6-a645-4376-bf64-512ebad60218.mov

francescospissu:
Nice rework👍 I tested it by comparing the behavior with RC-9 and it works as before. I also found the bug reported by @AlbyIanna but it was already present before these changes.

davegarthsimpson:
> The code looks fine to me overall, but I noticed a little bug.
> 
> It is possible, via shortcuts (CMD/CTRL + '+', CMD/CTRL + '-'), to exceed the min/max thresholds. For example, if the interface scale is set to `200` I can press CMD + '+' to increase the scale to `220`. When the threshold is exceeded, the stepper buttons might look enabled even if they should be disabled.

Thanks for catching this, this is actually because at the app level we are not restricting scale, we are only doing it with the stepper. This brings about the UX question: do we want to have a max/min scale. Either way the stepper itself is separate to this issue.

arduino-ide-extension/src/browser/dialogs/settings/settings-step-input.tsx
davegarthsimpson: yes! sorry I saw this after the fact, I'll add it in when we tackle the scale range question.

PR #1282, opened 2022-08-03
#1070 Can check if the current window is the first one.

### LINKED ISSUE #1070: [dev] As a dev I want to distinguish between first and other browser windows
### Describe the request

Why?

IDE2 can start with multiple browser windows when restoring all opened sketches. By default, the same frontend application loads in each browser window; hence the same logic runs. Running the same logic in each browser window is sometimes an undesired behavior. For example, when notifying the user about a new IDE2 version or asking the user to participate in a survey must happen in the first window.

The current code handles "the first window" workaround in different ways. See [here](https://github.com/arduino/arduino-ide/blob/f5cee97fefc1579693c1ceac853e41775cadb3d4/arduino-ide-extension/src/node/survey-service-impl.ts#L14-L16) and [here](https://github.com/arduino/arduino-ide/blob/f5cee97fefc1579693c1ceac853e41775cadb3d4/arduino-ide-extension/src/electron-main/ide-updater/ide-updater-impl.ts#L56-L59). This is a working solution, but adding a boilerplate service every time with a boolean flag can be generalized.

As a dev, I would expect to do something like this in the future:
```ts
@inject(WindowService) private readonly windowService: WindowService:

// later in code:
onStart(): void {
  this.windowService.isFirstWindow().then((first) => { if (first) { /* ask survey*/  }  );
}
```

In IDE2, we can customize the existing `WindowService` and extend it.

### Arduino IDE version

Version: 2.0.0-rc7 Date: 2022-06-09T08:19:19.877Z CLI Version: 0.23.0 [899dc91b]  Copyright © 2022 Arduino SA

### Operating system

macOS

### Operating system version

12.3.1

### Additional context

_No response_

### Issue checklist

kittaakos:
### Motivation

This is a dev thing. From now on, devs can call a service and check if the current window is the first one or not.

### Change description

New service for checking the state of the current window. See #1070 why this is needed.

How to verify?
 - Check out the code,
 - Modify any contributions to log whether or not the current window is the first one. Example:
```diff
diff --git a/arduino-ide-extension/src/browser/arduino-frontend-contribution.tsx b/arduino-ide-extension/src/browser/arduino-frontend-contribution.tsx
index f2b4af00..ac77421a 100644
--- a/arduino-ide-extension/src/browser/arduino-frontend-contribution.tsx
+++ b/arduino-ide-extension/src/browser/arduino-frontend-contribution.tsx
@@ -47,6 +47,7 @@ import { MonitorViewContribution } from './serial/monitor/monitor-view-contribut
 import { ArduinoToolbar } from './toolbar/arduino-toolbar';
 import { FrontendApplicationStateService } from '@theia/core/lib/browser/frontend-application-state';
 import { SerialPlotterContribution } from './serial/plotter/plotter-frontend-contribution';
+import { WindowServiceExt } from './theia/core/window-service-ext';
 
 @injectable()
 export class ArduinoFrontendContribution
@@ -78,6 +79,9 @@ export class ArduinoFrontendContribution
   @inject(FrontendApplicationStateService)
   private readonly appStateService: FrontendApplicationStateService;
 
+  @inject(WindowServiceExt)
+  private readonly windowServiceExt: WindowServiceExt;
+
   @postConstruct()
   protected async init(): Promise<void> {
     if (!window.navigator.onLine) {
@@ -89,6 +93,9 @@ export class ArduinoFrontendContribution
         )
       );
     }
+    this.windowServiceExt
+      .isFirstWindow()
+      .then((firstWindow) => console.log('is first window?', firstWindow));
   }
 
   async onStart(app: FrontendApplication): Promise<void> {
```
 - Start the IDE2, make sure you have at least two sketches opened,
 - Quit the IDE2,
 - Start the IDE2,
 - Once all windows are loaded, open the dev tools of the top most one, and it should show `true`,
 - Other windows should show `false`.

In action:

https://user-images.githubusercontent.com/1405703/182634668-5bc73548-3d70-41ce-8de9-204677d48cfd.mp4

### Other information
Closes #1070

### Reviewer checklist

### DIFF
--- a/arduino-ide-extension/package.json
+++ b/arduino-ide-extension/package.json
@@ -151,6 +151,10 @@
       "frontendElectron": "lib/electron-browser/theia/core/electron-menu-module"
     },
     {
+      "frontend": "lib/browser/theia/core/browser-window-module",
+      "frontendElectron": "lib/electron-browser/theia/core/electron-window-module"
+    },
+    {
       "electronMain": "lib/electron-main/arduino-electron-main-module"
     }
   ],
--- a/arduino-ide-extension/src/browser/contributions/first-startup-installer.ts
+++ b/arduino-ide-extension/src/browser/contributions/first-startup-installer.ts
@@ -43,7 +43,7 @@ export class FirstStartupInstaller extends Contribution {
             // If arduino:avr installation fails because it's already installed we don't want to retry on next start-up
             console.error(e);
           } else {
-            // But if there is any other error (e.g.: no interntet cconnection), we want to retry next time
+            // But if there is any other error (e.g.: no Internet connection), we want to retry next time
             avrPackageError = e;
           }
         }
@@ -64,7 +64,7 @@ export class FirstStartupInstaller extends Contribution {
             // If Arduino_BuiltIn installation fails because it's already installed we don't want to retry on next start-up
             console.log('error installing core', e);
           } else {
-            // But if there is any other error (e.g.: no interntet cconnection), we want to retry next time
+            // But if there is any other error (e.g.: no Internet connection), we want to retry next time
             builtInLibraryError = e;
           }
         }
--- a/arduino-ide-extension/src/browser/theia/core/browser-window-module.ts
+++ b/arduino-ide-extension/src/browser/theia/core/browser-window-module.ts
@@ -1,0 +1,10 @@
+import { DefaultWindowService as TheiaDefaultWindowService } from '@theia/core/lib/browser/window/default-window-service';
+import { ContainerModule } from '@theia/core/shared/inversify';
+import { DefaultWindowService } from './default-window-service';
+import { WindowServiceExt } from './window-service-ext';
+
+export default new ContainerModule((bind, unbind, isBound, rebind) => {
+  bind(DefaultWindowService).toSelf().inSingletonScope();
+  rebind(TheiaDefaultWindowService).toService(DefaultWindowService);
+  bind(WindowServiceExt).toService(DefaultWindowService);
+});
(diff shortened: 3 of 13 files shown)


AlbyIanna (approved):
The code's looking great! Thanks, @kittaakos 

I would just add one more thing to this PR:
Here below we are asking the electron backend to check for updates (it is done by every renderer process) 
https://github.com/arduino/arduino-ide/blob/bf193b1cac35a7dad1cfb6029541d14be1adfeb7/arduino-ide-extension/src/browser/contributions/check-for-updates.ts#L45
and the electron main process is smart enough to return early if the request was already been made by another renderer process so that in the end we only check for updates once.
https://github.com/arduino/arduino-ide/blob/bf193b1cac35a7dad1cfb6029541d14be1adfeb7/arduino-ide-extension/src/electron-main/ide-updater/ide-updater-impl.ts#L56-L59

But with your improvement, now we could avoid sending a request to electron every time a new renderer is spawn.

What do you think?

kittaakos:
Thank you for the review!

> But with your improvement, now we could avoid sending a request to electron every time a new renderer is spawn.
> 
> What do you think?

Exactly, excellent remark! This is my long-term plan, but I did not want to change the service APIs in this PR. I'd prefer handling them in a separate PR, removing the backend part of the survey service, and applying the IDE updater changes you have mentioned.

Eventually, I can make the changes here if you think it makes sense. What do you think?

AlbyIanna:
> but I did not want to change the service APIs in this PR. I'd prefer handling them in a separate PR

this sounds like a good plan, let's go for it 😉

francescospissu (approved):
Great work, thanks @kittaakos!

I have tried it and it works as expected. Thanks to this generalized behavior we can avoid creating a custom service implementation just to check if the current window is the first as I did here:
https://github.com/arduino/arduino-ide/blob/f5cee97fefc1579693c1ceac853e41775cadb3d4/arduino-ide-extension/src/node/survey-service-impl.ts#L13-L19

I will modify this implementation accordingly to your improvements.

kittaakos:
Thank you for the review @AlbyIanna, @francescospissu 🙏 

> I will modify this implementation accordingly to your improvements.

Sounds great. Let's tackle the API updates afterward.

PR #1337, opened 2022-08-19
Add tests for BoardsServiceProvider

AlbyIanna:
### Motivation
The following tests — marked with 1) and 2) — are failing on the main branch and should succeed here.

```
When there is one recognized boards attached
  and there is one unrecognized devicec available on a port
    and the port with the recognized board gets selected
      and then the port with the unrecognized board gets selected
        and then the port with the recognized board gets selected again
          and the recognized board gets unplugged because of an upload in progress
            1) board should remain selected with NO port selected
            and the recognized board gets plugged in again into the same port
              2) board should remain selected with NO port selected
```

### Reviewer checklist

### DIFF
--- a/arduino-ide-extension/src/browser/boards/boards-service-provider.ts
+++ b/arduino-ide-extension/src/browser/boards/boards-service-provider.ts
@@ -65,6 +65,13 @@ export class BoardsServiceProvider implements FrontendApplicationContribution {
   protected _availablePorts: Port[] = [];
   protected _availableBoards: AvailableBoard[] = [];
 
+  private uploadInProgress = false;
+
+  private lastItemRemovedForUpload: { board: Board; port: Port } | undefined;
+  // "lastPersistingUploadPort", is a port created during an upload, that persisted after
+  // the upload finished, it's "substituting" the port selected when the user invoked the upload
+  private lastPersistingUploadPort: Port | undefined;
+
   /**
    * Unlike `onAttachedBoardsChanged` this even fires when the user modifies the selected board in the IDE.\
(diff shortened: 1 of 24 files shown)


arduino-ide-extension/src/browser/boards/boards-service-provider.ts
```diff
@@ -364,12 +367,22 @@ export class BoardsServiceProvider implements FrontendApplicationContribution {
...
 
+  get latestValidBoardsConfig():
+    | RecursiveRequired<BoardsConfig.Config>
+    | undefined {
+    return this._latestValidBoardsConfig;
+  }
+
+  get latestBoardsConfig(): BoardsConfig.Config | undefined {
+    return this._latestBoardsConfig;
+  }
```
kittaakos, line 378: Do you use public APIs for the tests or do you think this will be used anywhere else in the app?

If the former, I propose using `protected` and subclass for the tests and increasing the visibility. If the latter, please explain where you plan to use it.

arduino-ide-extension/src/test/browser/boards-service-provider.test.ts
```diff
@@ -0,0 +1,311 @@
...
+              ports: [],
+            },
+            uploadInProgress: false,
+          });
+        });
+
+        it('should emit a onAvailablePortsChanged event', async () => {
+          subject.onAvailablePortsChanged(() => {
+            expect(true).to.be.ok;
+          });
```
kittaakos, line 143: Does this test fail if the event is not received?

AlbyIanna:
Closing this PR for now as the tested feature changed substantially since these tests were written.

PR #1479, opened 2022-09-21
Avoid intellisense widgets being covered by the bottom panel

r3inbowari:
### Motivation
Avoid the problem that the Intellisense widget may be cover by the bottom(Output) panel partially.

### Change description
from: 
image

to: 
image

### Reviewer checklist

### DIFF
--- a/arduino-ide-extension/src/browser/style/main.css
+++ b/arduino-ide-extension/src/browser/style/main.css
@@ -2,10 +2,14 @@
     background: var(--theia-editorGroupHeader-tabsBackground);
 }
 
-/* Negative values to lower the priority in order to avoid the problem that
-   the Intellisense widget may be cover by the bottom panel partially. */
-#theia-bottom-content-panel {
-   z-index: -1;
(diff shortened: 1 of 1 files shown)


AlbyIanna (changes requested):
Thanks for your contribution @r3inbowari!

Your change fixes the issue, although I would prefer increasing the `z-index` of the `#theia-main-content-panel` instead of setting a negative value for the `#theia-bottom-content-panel`. This way, we make sure that any sibiling of the main content panel will have a lower z-index than its own – for example, in the situation we add a new widget there, we would incur in the same issue again.

Something like this should work:
```
#theia-main-content-panel {
  z-index: 1
}
```

r3inbowari:
Agree :) Also fixed SplitPanel-handle showing above intellisense widget.

from:


to:

AlbyIanna:
This looks even better! I tested it and it works perfectly 😺 
Thank you so much!

AlbyIanna (approved):

per1234 (approved):
Thanks @r3inbowari!

PR #1490, opened 2022-09-23
Initialise the IDE updater even when 'checkForUpdates' preference is false

AlbyIanna:
### Motivation
"Check for Arduino IDE Updates" operation fails when automatic update checks are disabled #1437

The reason is that we're avoiding to initialise the updater properly at start-up if the "Check for Arduino IDE Updates" preference is disabled.

### Change description
Always initialise the updater. Only then we can decide if effectively check for updates or not.

### Reviewer checklist


kittaakos (approved):
It works for me. Thank you!

PR #1517, opened 2022-10-04
Fix status bar clipped in minimal state

r3inbowari:
### Motivation

Fix bottom status bar(or top tool bar when click the `output` title) is clipped by allowing the `DockPanel` has a wider height range.

### Change description

Make the total height of the split children less than the preset [minHeight](https://github.com/arduino/arduino-ide/commit/99664ee544d146c43a047547ea37e035eb90ff95) as much as possible.

### Other information

Related to #1515 #1151 #779 #662
### Reviewer checklist

### DIFF
--- a/arduino-ide-extension/src/browser/style/index.css
+++ b/arduino-ide-extension/src/browser/style/index.css
@@ -55,7 +55,8 @@
 /* Makes the sidepanel a bit wider when opening the widget */
 .p-DockPanel-widget {
   min-width: 200px;
-  min-height: 200px;
+  min-height: 20px;
+  height: 200px;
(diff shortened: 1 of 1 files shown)


per1234 (approved):
I verified it fixes https://github.com/arduino/arduino-ide/issues/1515 and https://github.com/arduino/arduino-ide/issues/662 when used with any reasonable combination of window size and scale values (they can still occur when the vertical window size is ridiculously small and the scale value ridiculously large, but the IDE UI in general will be unusable under those conditions so we don't need to be concerned about it).

Thanks @r3inbowari!

PR #1547, opened 2022-10-07
Let the user edit the font size settings with the keyboard

### LINKED ISSUE #1542: It‘s difficult to change the font size
### Describe the problem

The font size value is always locked at 8 or 72. 
Related to #1502.

### To reproduce

https://user-images.githubusercontent.com/30739857/194424678-63930371-0064-451d-b7e4-4d91a2f91e79.mp4

### Expected behavior

Adjustable size range via keyboard.

### Arduino IDE version

2.0.1-snaposhot-fe58e2b

### Operating system

Windows

### Operating system version

11

### Additional context

_No response_

### Issue checklist

AlbyIanna:
### Motivation
It's difficult for the user to edit the font size in the preferences using the keyboard.

On the current state, when a user tries to edit the value of the font size with the keyboard, the value would update unexpectedly.

For example:
- the font size is 72
- focus the input field
- press BACKSPACE
- expected: the font size is 7; result: the font size is 8 😢 
- now press 1
- expected: the font size is 81; result: the font size is 72 😢 

The reason is we validate the user input _as they type_, and if the input is invalid, we change the value to a valid one.

### Change description
With this change, I want to validate the user input _as they finished typing_ instead of _as they type_. In order to achieve that, I used the `onBlur` callback of the input field, so that the user can type whatever they want in the input field (only numbers, to be precise), and as soon as their focus change from the input field to another element (e.g.: they click on any other piece of UI or press ESCAPE) we validate the input they just typed in and, if necessary, change it to a valid value.

<details>
  <summary>Before the change (video):</summary>

https://user-images.githubusercontent.com/6939054/194602614-cc7f3ce5-b45e-49c5-83d5-ac0c553d7b9a.mov

</details>

<details>
  <summary>After the change (video):</summary>

https://user-images.githubusercontent.com/6939054/194602625-cf8fe2f3-ede7-4b6d-b791-9f5296f1c442.mov

</details>

### Other information
Fixes #1542 

### Reviewer checklist

### DIFF
--- a/arduino-ide-extension/src/browser/style/settings-step-input.css
+++ b/arduino-ide-extension/src/browser/style/settings-step-input.css
@@ -2,10 +2,10 @@
     position: relative
 }
 
-.settings-step-input-element::-webkit-inner-spin-button, 
+.settings-step-input-element::-webkit-inner-spin-button,
 .settings-step-input-element::-webkit-outer-spin-button {
-  -webkit-appearance: none;
-  margin: 0;
+    -webkit-appearance: none;
+    margin: 0;
 }
 
 .settings-step-input-buttons-container {
@@ -25,7 +25,7 @@
     right: 14px;
 }
 
-.settings-step-input-container:hover > .settings-step-input-buttons-container {
+.settings-step-input-container:hover>.settings-step-input-buttons-container {
     display: flex;
 }
 
--- a/arduino-ide-extension/src/browser/style/settings-step-input.css
+++ b/arduino-ide-extension/src/browser/style/settings-step-input.css
@@ -12,13 +12,17 @@
     display: none;
     flex-direction: column;
     position: absolute;
-    right: 14px;
+    right: 0px;
     top: 50%;
     transform: translate(0px, -50%);
     height: calc(100% - 4px);
     width: 14px;
     padding: 2px;
     background: var(--theia-input-background);
+}
+
+.settings-step-input-buttons-container-perc {
+    right: 14px;
 }
 
 .settings-step-input-container:hover > .settings-step-input-buttons-container {
--- a/arduino-ide-extension/src/browser/dialogs/settings/settings-component.tsx
+++ b/arduino-ide-extension/src/browser/dialogs/settings/settings-component.tsx
@@ -180,7 +180,8 @@ export class SettingsComponent extends React.Component<
           <div className="column">
             <div className="flex-line">
               <SettingsStepInput
-                value={this.state.editorFontSize}
+                key={`font-size-stepper-${String(this.state.editorFontSize)}`}
+                initialValue={this.state.editorFontSize}
                 setSettingsStateValue={this.setFontSize}
                 step={fontSizeStep}
                 maxValue={maxFontSize}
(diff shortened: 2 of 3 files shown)


davegarthsimpson (changes requested):
unfortunately there are a few issues here:

1. This breaks the arrow buttons functionality in general due to the wrong value being referenced in the `onStep` function
2. It results in a buggy feeling UX if the user clicks an arrow when the input value is '0' or '-' (the blur restores the initial value and then "steps it"), in the below changes I've made sure to disable the steppers when an invalid value is in the input
3. the value is passed as prop from the parent, when it changes it should reset the internal state, we should use the concept of an `initialValue` together with a key attribute in the stepper JSX to achieve this, the internal state does not need to be updated when `setSettingsStateValue` is invoked as doing should trigger a reset anyway

I pushed some changes to a different branch here which resolve the above:
https://github.com/arduino/arduino-ide/commit/ce3b47bad4b855b11a83f4d09c568a02c93254b7
There is also a small styling change as I noticed the position of the arrow buttons container was wrong in the font size stepper.

francescospissu:
It works well for me. The only small problem I encountered occurs when an invalid value is entered in the input field when the current value is the minimum (font size 8), in this case the input value is not automatically updated as entering an invalid value while the current value is greater than the minimum.

See the video:

https://user-images.githubusercontent.com/94986937/196705524-23a51dcf-e990-4425-bb80-e159eba1103c.mp4

AlbyIanna:
> It works well for me. The only small problem I encountered occurs when an invalid value is entered in the input field when the current value is the minimum (font size 8), in this case the input value is not automatically updated as entering an invalid value while the current value is greater than the minimum.

Nice catch! I just pushed a fix 👍

kittaakos:
I did not check the code changes thoroughly, but the behavior only. If I set the font size to 73 which exceeds 72 and set enter, the new value will be 72. If I set the value to 0, nothing changes. I expected symmetrical behavior; setting the value to 0 will set the font size to min; 8.

arduino-ide-extension/src/browser/style/settings-step-input.css
```diff
@@ -43,4 +47,4 @@
 
 .settings-step-input-button:hover {
     background: rgba(128, 128, 128, 0.8);
-}
+}
```
kittaakos, line 50: _No newline at the end of file_

davegarthsimpson (changes requested):
```
 const setValidValue = (value: number): void => {
    const clampedValue = clamp(value, minValue, maxValue);
    setValueState({ currentValue: clampedValue, isEmptyString: false });
    setSettingsStateValue(clampedValue);
  };
```

this doesn't feel right, if the value has changed `setSettingsStateValue` resets the whole component as well as the internal state, thus setting the internal state shouldn't be necessary

kittaakos (approved):
It works for me. Thank you!

AlbyIanna:
> this doesn't feel right, if the value has changed setSettingsStateValue resets the whole component as well as the internal state (because it changes the jsx key), thus setting the internal state shouldn't be necessary, if we're having to do this things are likely muddled and we probably need to rethink the lifecycle

Ok, I'll look into it better later 👍

AlbyIanna:
@francescospissu would you please give it another try? @davegarthsimpson just pushed a fix.

francescospissu (approved):
Now it works perfectly for me. Thanks @AlbyIanna 🙂

davegarthsimpson (approved):

PR #1758, opened 2022-12-09
Fix corruption of incoming UTF-8-encoded serial data

palazzol:
### Motivation
Fix for #589

### Change description
Fix corruption of incoming UTF8-encoded serial data

### Other information
Serial data is received in chunks.  When a UTF8-encoded symbol is multibyte, it cannot be decoded properly until all the bytes are available.  If an attempt at decoding is made, it will fail and the symbol for UTF-8 "Replacement Character" � is inserted into the stream.  TextDecoder has a "stream" option to buffer any bytes that cannot be decoded yet, for exactly this reason.  This fix creates an instance of TextDecoder which is persistent for the life of the connection, and then uses that decoder with the stream option to decode UTF8-encoded data properly.

Note - this assumes we expect UTF8 encoded data from the serial port.  Generally speaking, this happens if we have the sketch editor in UTF8 mode, which appears to be the default, and then UTF8 encoded characters are included in a print() statement.

### Reviewer checklist

### DIFF
--- a/arduino-ide-extension/src/node/monitor-service.ts
+++ b/arduino-ide-extension/src/node/monitor-service.ts
@@ -80,7 +80,7 @@ export class MonitorService extends CoreClientAware implements Disposable {
   private readonly board: Board;
   private readonly port: Port;
   private readonly monitorID: string;
-  private streamingTextDecoder = new TextDecoder('utf8');
+  private readonly streamingTextDecoder = new TextDecoder('utf8');
 
   /**
    * The lightweight representation of the port configuration currently in use for the running monitor.
--- a/arduino-ide-extension/src/node/monitor-service.ts
+++ b/arduino-ide-extension/src/node/monitor-service.ts
@@ -80,6 +80,7 @@ export class MonitorService extends CoreClientAware implements Disposable {
   private readonly board: Board;
   private readonly port: Port;
   private readonly monitorID: string;
+  private streamingTextDecoder = new TextDecoder('utf8');
 
   /**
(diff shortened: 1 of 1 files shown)


arduino-ide-extension/src/node/monitor-service.ts
```diff
@@ -80,6 +80,7 @@ export class MonitorService extends CoreClientAware implements Disposable {
   private readonly board: Board;
   private readonly port: Port;
   private readonly monitorID: string;
+  private streamingTextDecoder = new TextDecoder('utf8');
```
z3bra5hax, line 83:
```suggestion
  private readonly streamingTextDecoder = new TextDecoder('utf8');
```

z3bra5hax: Just to follow convention in reality, no reason this should be modified so let's just update this.

z3bra5hax (changes requested):

arduino-ide-extension/src/node/monitor-service.ts
palazzol: Makes sense to me, thanks

z3bra5hax (approved):

kittaakos (approved):
This is a quality contribution.

The code looks great and #589 works with the changes.

Thank you!

per1234 (approved):
I verified this fixes https://github.com/arduino/arduino-ide/issues/589

Thanks @palazzol!

PR #1893, opened 2023-02-18
feat: Exception Stack Trace Decoder for esp32

radurentea:
Exception Stack Trace Decoder for esp32 boards with emphasised data output

### Motivation
Decoding exception stack trace

### Change description
Adds a component component similar stylistically with the serial monitor in which the user can paste the backtrace.

https://user-images.githubusercontent.com/25748706/223938868-acd6f63d-2cfe-4233-9ad9-9bb4d454f5dd.mp4

### Other information
Based on the decoder from the old arduino ide (https://github.com/me-no-dev/EspExceptionDecoder)

### Reviewer checklist


kolipakakondal:
Hi @radurentea Could you add a screen recording to get a glimpse of this feature?

ubidefeo:
hi @radurentea 

We've been discussing this internally for a bit and came to the conclusion that such a feature should not be part of the stock IDE 2, but it should be provided when a user decides to install it, or even better as an extension to the ESP32 platform.

As we try to find ways for developers to access useful features in IDE 2.0, we've investigated the VSIX way, since Theia supports VS Code extension in that format.
Unfortunately such layer won't be able to access things which are very useful.
You want to know which board is selected, installed cores etc, and since extensions won't communicate with the CLI daemon there are a lot of limitations towards implementing this feature.

I have had a chat with @kittaakos regarding this, but maybe this PR can start a conversation and maybe an RFC.

I know for a fact that we don't have the time right now to implement a 3rd party usable API, but it's something which has been on our plate for a while and we really see a great amount of value.
I personally would love to see Stack Decoder in IDE 2 :) 

I'd also invite @alranel to chime in on the discussion.

frank26080115:
@radurentea can this be installed as some sort of overlay on top of existing IDE installations? As in, a set of files in a folder structure that the user has to simply copy and paste into an existing IDE installation directory? I looked at https://github.com/arduino/arduino-ide/pull/1893/commits/ed664791393365233092c912d5f60ccaf925fafb and `arduino-ide-frontend-module.ts` and `arduino-ide-frontend-module.js` and it seems like this is the only file where you made insertions, the rest are new files.

Perhaps it isn't too bad having the user making a 17 line patch into one JS file to enable this feature? Maybe later a wizard can be created.

kittaakos:
> can this be installed as some sort of overlay on top of existing IDE installations?

I am not aware of any easy way of doing it. Also, it won't work with the current code base anymore, as the frontend code is calling server code. [Here](https://github.com/arduino/arduino-ide/pull/1893/files#diff-d76f4b1dceaca33536c9e100d81d44cbc6e5cbdd5fd8d05dbb230619a919ec26R4) is an example.

radurentea:
Hi @dankeboy36,

I've made the decoder as a prototype, unfortunately, it only works for the ESP32 modules. The ESP8266 have a different format.

I could add more logic and add the regex that is needed to work for ESP8266 as well, but I'm not sure how soon I could do that because currently I don't have my hands on that chip.

If you want to have a try at it yourself, you can check the Exception Decoder from the old IDE. This is what I used as a reference  https://github.com/me-no-dev/EspExceptionDecoder/blob/master/src/EspExceptionDecoder.java

Just let me know!

dankeboy36:
> I've made the decoder as a prototype, unfortunately, it only works for the ESP32 modules.

Hi, @radurentea! I tried it with my ESP32-S2 Pico after installing the core from `https://raw.githubusercontent.com/espressif/arduino-esp32/gh-pages/package_esp32_index.json`. It did not work for me with the same sketch content. Does this prototype work with an ESP32-S2 board?

Veriy and upload worked. The monitor prints nothing, and the board connection is permanently lost. Can somebody please help? I want to try out this example to understand how it works. Thank you!

https://github.com/arduino/arduino-ide/assets/111981763/6f7c58d9-3af4-4617-bc0d-522a77366597

ubidefeo:
I am also unable to get the board to dump the stack after setting Debug Level to Verbose
any hint on why or how?

@dankeboy36  and I are both using an ESP32-S2 by Waveshare

kittaakos:
Hello, thank you for your contribution. I will close this PR without accepting it. Please use https://github.com/dankeboy36/esp-exception-decoder instead with today's nightly IDE2 from https://www.arduino.cc/en/software#nightly-builds. It's a drop-in Arduino IDE extension that you can use. If you happen to have any issues, please report them at https://github.com/dankeboy36/esp-exception-decoder/issues/new.

If you do not agree with something, please ping me.

PR #1913, opened 2023-02-22
Enhance board config auto-selection with hardwareId

davegarthsimpson:
### Note
This PR builds on [1885](https://github.com/arduino/arduino-ide/pull/1885), but its considered fundamentally different as it changes the UX for many other boards.

### Motivation
It may be convenient for users if the IDE2 were to "auto-select" a "new" port on their behalf in the following scenario: _when they select a board and port, unplug that board from its port, and plug it into a "new" port._

Now that we receive `hardwareId`s  from the CLI for boards that expose them, we can reliably "auto-select" a board/port in the above scenario, as we can distinguish between two boards with the same `fqbn`.

At present in the unplug/replug scenario we only auto-select a board-port combo when it's unplugged/plugged from/into the **same** port, or if a port change occur **during upload**.

### Change description
If a specific board's port changes to a "new" port, and that board exposes a `hardwareId`, its "new" port should always be selected automatically.
 
These changes are expected to enhance the UX for boards that expose a `hardwareId`, the UX will not change for boards that do not, for example in my (@davegarthsimpson) testing the Arduino Leonardo does not expose one.

#### Boards with hardwareId
-Their port should be automatically re-selected when it changes (while the board remains selected),
1. due to unplugging it from a port and replugging it into a **different port**;
2. due to unplugging it from and replugging it into the **same port**;
3. due to an upload, after upload is complete;

#### Board without hardwareId
-Their port will only be re-selected when it changes (while the board remains selected), 
1. due to unplugging it from and replugging it into the **same port**;
2. due to an upload, after upload is complete;

Example boards with hardwareId (to the authors knowledge): Arduino Zero, Arduino Uno
Example boards without hardwareId (to the authors knowledge): Arduino Leonardo, Arduino Due (Native USB Port)

### Reviewer checklist

* [x] Docs have been added / updated (for bug fixes / features)


arduino-ide-extension/src/browser/boards/boards-service-provider.ts
```diff
@@ -205,20 +199,36 @@ export class BoardsServiceProvider
...
-        )
-      ) {
+        );
+
+        const boardIsSameFqbn = Board.sameAs(
+          boardOnAppearedPort,
+          lastBoardsConfigOnUpload.selectedBoard
+        );
+
+        if (!boardIsSameHardware && !boardIsSameFqbn) return;
```
kittaakos, line 215: Instead of `return`, it should be `continue`, right?

kittaakos:
> This PR builds on https://github.com/arduino/arduino-ide/pull/1885,

If anybody is wondering what's the effective code change: https://github.com/arduino/arduino-ide/compare/include-hardwareId-in-auto-selection...enhance-auto-selection-with-hardwareId

arduino-ide-extension/src/browser/boards/boards-service-provider.ts
davegarthsimpson: this is the same as in [1885](https://github.com/arduino/arduino-ide/pull/1885), and should result in the same reasoning as 2.0.3, but considering `hardwareId`: if a board that "appeared" does not have the same `hardwareId` or `fqbn` as the one selected for upload, don't do anything and return.

davegarthsimpson: also to clarify this logic is only related to port change during upload

kittaakos: No. Without this change, IDE2 goes through all appeared ports and tries to find the matching one. It's a noop, when a match was found. With your changes, the logic breaks after the first non-match.

davegarthsimpson: for context, in 2.0.3 we have [this](https://github.com/arduino/arduino-ide/blob/01ee045beb1cd69b23bc68c83b918bb14ab6ef91/arduino-ide-extension/src/browser/boards/boards-service-provider.ts#L210), the "if block" is avoided if the "appeared board" is not the same fqbn, so same as doing a return as above (but it also considers `hardwareId`)

kittaakos (commented):
I had one remark regarding the `name` fallback. The rest look OK to me. Thanks

arduino-ide-extension/src/browser/boards/boards-service-provider.ts
```diff
@@ -205,20 +199,36 @@ export class BoardsServiceProvider
...
+
+        if (!boardIsSameHardware && !boardIsSameFqbn) continue;
+
+        let boardToAutoSelect = boardOnAppearedPort;
+        if (boardIsSameHardware && !boardIsSameFqbn) {
+          const { name, fqbn } = lastBoardsConfigOnUpload.selectedBoard;
+
+          boardToAutoSelect = {
+            ...boardToAutoSelect,
+            name: boardToAutoSelect.name || name,
```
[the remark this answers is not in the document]
davegarthsimpson: name and fqbn can be unavailable if a board is in DFU mode

per1234 (approved):
It is working perfectly for me. Thanks Dave!

To anyone interested in testing this: I'll share the procedure I found:

### Required Equipment

- A board with the following characteristics:
  - Persistent bootloader mode activated by a double reset
  - Both the application port and the bootloader port have the same hardware ID (usually a serial number)
- A Windows or macOS PC
  - Linux is not suitable because it doesn't assign different addresses to the application and bootloader ports. 

I found that the [**Portenta H7**](https://docs.arduino.cc/hardware/portenta-h7) was the only board I had on hand with the required characteristics.

### Steps

1. Connect the board to your computer.
1. Select the board's port in Arduino IDE.
1. Press and release the reset button on the board twice quickly to put it in bootloader mode.

Before the change made by this PR, no port would be selected after the board's port address changed when it went into bootloader mode.

After the change made by this PR, the port selection automatically follows the board to the address of the bootloader port.

davegarthsimpson (commented):
I have also done manual tests of this PR, involving some testing some hard-coded changes to simulate everything "working as it should" in the process chain.

From my tests, the board selector correctly auto-selects the correct port with hardwareId, when a board exposing a hardwareId disappears from a port and reappears on a different port in a different app state.

kittaakos:
This PR has been created to support the "auto-reconnecting" to a board in DFU mode, which the R4 Minima goes into during upload and reset. The primary motivation was to make IDE2 compatible with the Uno R4 board.

PR #875, opened 2022-02-28
handling pagination in getting the sketches

ulemons:
### Motivation

The remote sketchbook is limited to 20 sketches. It is needed to enable the pagination in order to get all the remote sketches.

### Change Description

Instead of doing just one call to retrive all the sketches, the component will keep calling the api with a window of 50 sketches per call. Untill it will get all of them.

### Reviewer checklist

- [ ] PR address a single concern.
- [ ] The PR has no duplicates (please search among the [Pull Requests](https://github.com/arduino/arduino-ide/pulls) before creating one).
- [ ] PR title and description are properly filled.
- [ ] Docs have been added / updated (for bug fixes / features)

PR #2014, opened 2023-04-12
fix: reduced unnecessary `GET /sketches` request

kittaakos:
### Motivation

This PR slightly modifies the implementation of [arduino/arduino-ide#875](https://github.com/arduino/arduino-ide/pull/875) to reduce the unnecessary `GET /sketches`. The endpoint claims to support pagination, but it does not. So if the number of sketches can be divided by the `limit` without a remainder, another fetch is required.

This PR does not change the limit but leaves it at 50. This PR adds tests to verify the new behavior. Note that those tests do not run on the CI, only locally when secrets are set in the environment.

To verify:
 - you have more than 50 sketches on https://create.arduino.cc/editor,
 - open IDE2 and log in to the cloud, and
 - verify that you see all you sketches in the cloud sketchbook tree.

### Reviewer checklist

### DIFF
--- a/arduino-ide-extension/src/browser/create/create-api.ts
+++ b/arduino-ide-extension/src/browser/create/create-api.ts
@@ -86,26 +86,25 @@ export class CreateApi {
     url.searchParams.set('user_id', 'me');
     url.searchParams.set('limit', limit.toString());
     const headers = await this.headers();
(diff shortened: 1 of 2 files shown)


per1234 (approved):
Nice refactoring Akos!

Issue #289, opened 2021-03-29
Tear off serial monitor window (and maybe others)


Lorandil:
In IDE 2.0 the serial monitor window is located at the bottom of the main window.
I like to keep my code window clean, so with IDE 1.8.13 I move the serial monitor to the second screen.
This allows me to watch outputs while making changes to the code without wasting screen space.
The functionality is missing in IDE 2.0 beta4.

It would probably make sense for other windows (output), too.

ag88:
a tear off serial monitor is rather useful, but i'd think a 'better' option is to delegate it to a dedicated serial monitor app like putty
https://www.putty.org/
good serial monitors with a lot of features take a lot of effort to develop. it would be better to find existing solutions that already works well
a problem though is for a 'serial monitor' that integrates with a graphing tool. that i'd guess could be met with a dedicated tool as well.

Lorandil:
Thanks for the tip.
The problem with external tools like putty is, that they block the serial port.
The IDE won't be able to upload to the micro controller until the connection is manually closed.
With the integraded serial monitor the IDE can claim and release the port as necessary.

Xtreemtec:
Started today with 2.0 and this was the first thing i missed..   We run a triple screen setup for programming and our debug is always on a second screen as we normally have a lot of data coming from our processors while programming. 
Also that CTRL + ENTER to send is very annoying..   But i will look if there is an issue open for that too.

ironhal0:
I started with 2.0 today, and this was also the first thing that I missed. Any apps that don't let me tear off individual functions to separate screens are essentially hobbling my 3 monitor setup and I take a big step backward in efficiency.  Being able to watch the serial output in its own large window while keeping code open in a separate large window is essential.

DaveTur:
It would also be very useful to 'Tear off' or 'float' the Debugger area.  Being able to tear off the individual debugging windows such as CALL STACK and WATCH would be even better.

PerryBebbington:
+1 for this, for all the reasons others have already given.

tomjennings:
I'm a heavy user of serial monitor. 1.8.19, vs 2.0.0 here. My projects are sometimes large with voluminous runtime output during development phase. Example below. "Tear off" or somehow separate window is critically important, especially when I'm field-debugging with a laptop and small screen. 

[...]

PR #2102, opened 2023-06-17
feat: can dock the monitor widget to the `"right"`

dankeboy36:
### Motivation

This is a quality of life improvement until the _Serial Monitor_ widget is not detachable from the IDE2 `window`: [arduino/arduino-ide#289](https://github.com/arduino/arduino-ide/issues/289). This is the simplest solution to use larger screens more efficiently. One can keep the widget at the `"bottom"`, or move it to the `"right"`.

https://github.com/arduino/arduino-ide/assets/111981763/2c08ab81-e266-4ba1-b209-8dead5fe1217

### Change description

Added a new preference (`arduino.monitor.dockArea`) to specify the area of the application shell where the _Serial Monitor_ widget resides. The possible values are `"bottom"` and `"right"`. The default value is `"bottom"`.

The dock area is per application and not per workspace or window. However, advanced users can create the `./.vscode/settings.json` and configure it per sketch preference.

### Reviewer checklist


kittaakos (commented):
`monitor.widgetDockArea`?

arduino-ide-extension/src/browser/arduino-preferences.ts
```diff
@@ -258,6 +266,17 @@ export const ArduinoConfigSchema: PreferenceSchema = {
       ),
       default: undefined,
     },
+    'arduino.monitor.dockArea': {
+      type: 'string',
+      enum: ['bottom', 'right'],
```
kittaakos, line 271: Maybe translate the enums. Let's check how IDE2 and Theia behave with enum preference values.

per1234 (commented):
I'm wondering about the term "area". I have struggled to understand the correct terminology to use when referring to Arduino IDE UI components.

Here we have the terms "Side Bar" for the UI containers on the left and right side of the window and and "Panel" for the container at the bottom:

- https://code.visualstudio.com/docs/getstarted/userinterface
- https://code.visualstudio.com/api/ux-guidelines/overview

Here we have the terms "area" and "panel" used for all three containers (but evidently not for the "main area"?):

https://www.typefox.io/blog/flexible-window-layout-in-theia-ide

I don't like how VS code uses a different term for a container depending on its location. I think it is more intuitive (especially in the context of Theia's "flexible window layout") to use the same term for all containers. I find "area" to be overly ambiguous, so I have been using the term "panel":

- "main panel"
- "bottom panel"
- "left panel"
- "right panel"

arduino-ide-extension/src/browser/arduino-preferences.ts
kittaakos: > Maybe translate the enums.

No. Other enum values are not translated in the _Advanced settings_ UI either. IDE2 can support translations in the IDE2 preferences dialog.

Screenshot 2023-06-21 at 13 58 24

per1234:
I've thought about it more and I think that in general it is best to use the terminology established by VS Code since this is by far a more significant precedent and more likely to already be familiar to the users than the alternate terminology established (inconsistently) in an obscure blog post.

However, it is not possible to follow the VS Code terminology exactly in this case because here we need a single term that can apply to both the "Panel" and the "Secondary Sidebar". Obviously the term "sidebar" can't be used for that purpose, so we are left with "panel".

So I suggest renaming the setting ID `arduino.monitor.dockPanel`. This will cause the setting name shown in the GUI advanced settings interface to be "**Arduino > Monitor: Dock Panel**" instead of the current "**Arduino > Monitor: Dock Area**"

image

The existing "bottom" and "right" setting values will work fine in that context. I think the setting description is OK as it is now.

per1234:
@dankeboy36 is the PR ready for another review round?

dankeboy36:
> PR ready for another review round?

Yes. Please take another look at the changes. The new preference key is now `arduino.monitor.dockPanel`. I have updated the PR description too. Thank you!

arduino-ide-extension/src/browser/arduino-preferences.ts
```diff
@@ -258,6 +266,17 @@ export const ArduinoConfigSchema: PreferenceSchema = {
       ),
       default: undefined,
     },
+    'arduino.monitor.dockArea': {
```
per1234, line 269:
```suggestion
    'arduino.monitor.dockPanel': {
```

per1234 (changes requested):

arduino-ide-extension/src/browser/arduino-preferences.ts
dankeboy36: Excellent catch. The compiler should find such errors, so I [added](https://github.com/arduino/arduino-ide/pull/2102/files#diff-9ffaadbfa5cea6de54ac86a7b454b1217c95e811a98186c11af35c783f59015dR57-R61) a tiny TypeScript trick to let the compiler error when there is a typo next time.

per1234 (approved):
Thanks @dankeboy36!

Issue #2112, opened 2023-06-23
Start the CLI daemon process without a shell


kittaakos:
### Describe the problem

This is a follow-up task of https://github.com/arduino/arduino-ide/pull/2067:

The CLI daemon process is a leftover:

https://github.com/arduino/arduino-ide/blob/d79bc0d083c40b3244264696d2d6703756a8bd53/arduino-ide-extension/src/node/arduino-daemon-impl.ts#L174-L175

### To reproduce

It's a dev thing.

### Expected behavior

Processes are not started with a shell.

### Arduino IDE version

d79bc0d

### Operating system

macOS

### Operating system version

13.4

#### Related

- https://forum.arduino.cc/t/cannot-start-arduino-2-1-0-as-normal-user/1135289/4
- https://forum.arduino.cc/t/arduino-ide-2-1-0-only-pulsing-arduino-symbol-after-launching-the-app/1121605/29
- https://forum.arduino.cc/t/cant-open-the-arduino-ide-2-1-1/1151978
- https://forum.arduino.cc/t/major-installation-bug-november-2022-windows-11/1051207
- https://forum.arduino.cc/t/arduino-2-1-1-wont-start/1151538
- https://forum.arduino.cc/t/arduino-ide-v2-hangs-at-start-up/1030052
- https://forum.arduino.cc/t/cannot-open-arduino-ide/1114087
- https://forum.arduino.cc/t/i-have-just-installed-ide-2/1022730
- https://github.com/arduino/arduino-ide/issues/2073

### Issue checklist

- [X] I searched for previous reports in [the issue tracker](https://github.com/arduino/arduino-ide/issues?q=)
- [X] I verified the problem still occurs when using the latest [nightly build](https://www.arduino.cc/en/software#nightly-builds)
- [X] My report contains all necessary details

PR #2182, opened 2023-08-20
fix: execute the Arduino CLI without a `shell`

kittaakos:
### Motivation

Execute the Arduino CLI process without a `shell`.
 - security reasons,
 - some users might not have the `%ComSpec%` set on Windows, and `CMD.exe` is not on the `$PATH`

Remove the obsolete CLI daemon watcher process. Before https://github.com/arduino/arduino-cli/pull/488, IDE2 required a way to stop the daemon process if the parent (backend) process crashed. However, this mechanism
is no longer necessary, as the CLI daemon process is not actually a true daemon process.

### Other information

To verify:
 - Arduino IDE starts, and the CLI is fully functional when there is a whitespace in the path to your IDE2 installation.
 - Start IDE2. Open the task manager/activity monitor on your system. See that `arduino-cli` process is running. Kill the IDE2 process. See that the `arduino-cli` process is not running.

Closes https://github.com/arduino/arduino-ide/issues/2112
Ref: https://github.com/arduino/arduino-ide/pull/2067

### Reviewer checklist


per1234 (approved):
I verified this fixes the problem with Arduino failing to start up with errors like:

```text
daemon ERROR Error: spawn C:\WINDOWS\system32\cmd.exe ENOENT
    at Process.ChildProcess._handle.onexit (node:internal/child_process:282:19)
    at onErrorNT (node:internal/child_process:480:16)
    at processTicksAndRejections (node:internal/process/task_queues:83:21)
```

in cases where there is a problem with the user's `ComSpec` environment variable:

- Variable is not defined ([example](https://forum.arduino.cc/t/arduino-ide-v2-hangs-at-start-up/1030052/6))
- Variable is set to an incorrect path ([example](https://forum.arduino.cc/t/cant-open-the-arduino-ide-2-1-1/1151978/12))
- `cmd.exe` has been removed ([example](https://forum.arduino.cc/t/major-installation-bug-november-2022-windows-11/1051207/8))
- Variable has been set to the path of an incompatible shell ([example](https://forum.arduino.cc/t/arduino-ide-2-1-0-only-pulsing-arduino-symbol-after-launching-the-app/1121605/29))

Thanks Akos!

PR #2457, opened 2024-06-17
feat: use Arduino CLI 1.0.4

giacomocusinato:
### Motivation

Use Arduino CLI `1.0.0` APIs in IDE2.

### Change description

Update CLI version to `1.0.0`, bind new APIs

### Reviewer checklist


arduino-ide-extension/src/node/config-service-impl.ts
```diff
@@ -182,7 +182,11 @@ export class ConfigServiceImpl
       });
       const model = (yaml.safeLoad(content) || {}) as DefaultCliConfig;
       this.logger.info(`Loaded CLI configuration: ${JSON.stringify(model)}`);
-      if (model.directories.data && model.directories.user) {
+      if (
+        model.directories &&
```
dankeboy36, line 186: If `directories` is optional, I propose reflecting it on the type:
https://github.com/arduino/arduino-ide/blob/2e78e96b75b8c8c3971f6b115eec558ba84d0317/arduino-ide-extension/src/node/cli-config.ts#L98

arduino-ide-extension/src/node/config-service-impl.ts
```diff
@@ -217,13 +221,20 @@ export class ConfigServiceImpl
 
   private async getFallbackCliConfig(): Promise<DefaultCliConfig> {
     const cliPath = this.daemon.getExecPath();
-    const rawJson = await spawnCommand(cliPath, [
+    const rawJson = await spawnCommand(cliPath, ['config', 'dump', '--json']);
+    const config = JSON.parse(rawJson);
+
+    // Since CLI 1.0, the command `config dump` only returns user-modified values and not default ones.
+    // directories.user and directories.data are required by IDE2 so we get the default value explicitly.
+    const directoriesRaw = await spawnCommand(cliPath, [
```
dankeboy36, line 229: The CLI executions do not depend on each other: they can spawn parallel with `Promise.all()`

arduino-ide-extension/src/node/config-service-impl.ts
```diff
@@ -291,48 +302,64 @@ export class ConfigServiceImpl
...
+      'directories.user': config.directories.user,
+      'directories.data': config.directories.data,
+      'network.proxy': config.network?.proxy,
+      'board_manager.additional_urls':
+        config.board_manager?.additional_urls || [],
+    };
+
+    for (const [key, value] of Object.entries(updatableConfig)) {
+      const client = createArduinoCoreServiceClient({ port });
+      const req = new SettingsSetValueRequest();
```
dankeboy36, line 322: Is this a new CLI behavior? I see a client is created before setting each value. Can it be moved outside of the loop?

arduino-ide-extension/src/node/config-service-impl.ts
```diff
@@ -291,48 +302,64 @@ export class ConfigServiceImpl
...
+          resolve(resp.getEncodedSettings());
         } finally {
           client.close();
         }
       });
     });
+
+    const cliConfigUri = await this.getCliConfigFileUri();
+    const cliConfigPath = FileUri.fsPath(cliConfigUri);
+    fs.writeFile(cliConfigPath, configRaw, { encoding: 'utf-8' });
```
dankeboy36, line 362: missing `await` or `return`

arduino-ide-extension/src/node/config-service-impl.ts
giacomocusinato: Yes CLI config API's were refactored and they now only exposed methods for retrieving/setting single values, instead of the previous "merge" approach. 
I agree the client should be created one time outside the loop. Thanks for the review! 👍

arduino-ide-extension/src/node/config-service-impl.ts
giacomocusinato: I think `directories` can be undefined only on the initial parse of the current config. 
We can manipolate the loaded config  object to assure the directories are alway defined  object returned by `loadCliConfig` is always of type `DefaultCliConfig`.

That's the current logic: 

1. Load current cli config as type `CliConfig` (optional `directories`).
2. Check `directories` value:
    1. If defined, return config as `DefaultCliConfig` type
    2. If undefined, call `getFallbackCliConfig` to get default directories, merge them to current config and return the object as `DefaultCliConfig`

davegarthsimpson (approved):

Issue #2547, opened 2024-10-30
Modification of installed core files breaks compilation


giacomocusinato:
### Describe the problem

When IDE is open, modifying installed core files (like `boards.txt` or `programmers.txt`) leads to compilation not working with the following error logged in the console: `The instance is no longer valid and needs to be reinitialized`

### To reproduce

- Open IDE
- Select any board (for example Arduino Nano, but any will do)
- Locate `Arduino15` folder on your OS https://support.arduino.cc/hc/en-us/articles/360018448279-Open-the-Arduino15-folder
- Locate `board.txt` file relative to the board you selected (for Arduino Nano is `Arduino15/packages/arduino/hardware/avr/1.8.6
`)
- Modify the file `board.txt` in any way
- Compile the sketch 

The error `The instance is no longer valid and needs to be reinitialized` should appear

### Expected behavior

The compilation should proceed as expected as the IDE should handle the error silently.

### Arduino IDE version

2.3.3

### Operating system

macOS

### Operating system version

15.0.1

### Issue checklist

- [x] I searched for previous reports in [the issue tracker](https://github.com/arduino/arduino-ide/issues?q=)
- [x] I verified the problem still occurs when using the latest [nightly build](https://www.arduino.cc/en/software#nightly-builds)
- [x] My report contains all necessary details

PR #2548, opened 2024-10-30
fix: retry compilation if grpc client needs to be reinitialized  

giacomocusinato:
### Motivation

#2547 

### Reviewer checklist


arduino-ide-extension/src/node/core-service-impl.ts
```diff
@@ -112,6 +112,15 @@ export class CoreServiceImpl extends CoreClientAware implements CoreService {
             reject(UserAbortApplicationError());
             return;
           }
+
+          if (
+            ServiceError.isInvalidArgument(error) &&
+            error.details.includes('instance is no longer valid')
```
dankeboy36, line 118: What should happen when the selected language is, for example, German, and there is a translation for this error message for the CLI?

dankeboy36 (commented):
It's looking great.

Why not handle the error on the backend without sending it back to the FE? Is there a technical limitation? Thank you! (Related https://github.com/arduino/arduino-ide/pull/2500)

arduino-ide-extension/src/node/core-service-impl.ts
giacomocusinato: I'm now extracting the error type from the error metadata 👍

giacomocusinato:
Thanks for the review @dankeboy36!
You're right, would be best to handle the error directly on the backend. Initially I found a bit confusing code-wise restarting the compile stream in the same method, but I've proposed a solution now. If that works I might use the same approach  with the logic in https://github.com/arduino/arduino-ide/pull/2500

davegarthsimpson (approved):

PR #2581, opened 2024-11-26
feat: introduce donation links in update flow

giacomocusinato:
### Motivation

- Add footer to `IDEUpdaterDialog` with donate link
- Show donate dialog if first instance of a new version of the IDE is running

### Reviewer checklist


arduino-ide-extension/src/browser/dialogs/version-welcome-dialog.tsx
```diff
@@ -0,0 +1,93 @@
...
+        <p>
+          {nls.localize(
+            'arduino/versionWelcome/donateMessage',
+            'Arduino is committed to keeping software free and open-source for everyone. Your donation helps us develop new features, improve libraries, and support millions of users worldwide.'
+          )}
+        </p>
+        <p className="bold">
+          {nls.localize(
+            'arduino/versionWelcome/donateMessage2',
+            'Please consider supporting our efforts to keep Arduino IDE free.'
```
per1234, line 43:
```suggestion
            'Please consider supporting our work on the free open source Arduino IDE.'
```

The previous message implies there is a plan to make the project non-free. That is not an appropriate tone to take in our communications to the users.

per1234 (changes requested):

arduino-ide-extension/src/browser/contributions/check-for-ide-updates.ts
```diff
@@ -66,4 +85,26 @@ export class CheckForIDEUpdates extends Contribution {
...
+    }
+  }
+
+  private async isRunningNewIDEVersion(): Promise<boolean> {
+    const { appVersion } = await this.appService.info();
+    const prevVersion = await this.localStorage.getData<string>(
+      LAST_USED_IDE_VERSION
+    );
+    try {
+      return !!prevVersion && new SemVer(appVersion).compare(prevVersion) === 1;
```
per1234, line 105:
```suggestion
      const appVersionObject = new SemVer(appVersion);
      return (
        !!prevVersion &&
        appVersionObject.prerelease.length === 0 &&
        appVersionObject.compare(prevVersion) === 1
      );
```

I think it is better to suppress the dialog when the user is using a pre-release version. It is sufficient to show it only after updates to production releases.

The previous code will cause the appearance of the dialog to occur intermittently for users of the [tester builds](https://github.com/arduino/arduino-ide/blob/main/docs/contributor-guide/beta-testing.md#testing-pull-requests), since these builds use the format `X.Y.Z-snapshot-<commit hash>`, and a tester build version may be considered higher or lower than another version at random purely based on the [lexicographical relationship](https://semver.org/#spec-item-11) between the two commit hashes. This apparent inconsistency might cause beta testers to think there is a bug in the system.

The previous code will cause the appearance of the dialog every time the user updates to a new [nightly build](https://github.com/arduino/arduino-ide/blob/main/docs/contributor-guide/beta-testing.md#testing-nightly-build) (which a dedicated beta tester will do frequently).

The prerelease builds are primarily used by beta testers. [Beta testing is an valuable contribution](https://github.com/arduino/arduino-ide/blob/main/docs/contributor-guide/beta-testing.md#beta-testing-guide) already so I don't think it is fair to subject these users to more excessive donation requests than to those who only use release builds.

I'm not skilled in writing TypeScript so the code I suggest here might be suboptimal. I provide it more to clearly define the nature of my proposal. Feel free to implement the code anyway you like if my proposal is accepted.

arduino-ide-extension/src/browser/dialogs/version-welcome-dialog.tsx
```diff
@@ -0,0 +1,99 @@
...
+    );
+    cancelButton.classList.add('secondary');
+    cancelButton.classList.add('cancel-button');
+    this.addAction(cancelButton, this.close.bind(this), 'click');
+    this.controlPanel.appendChild(cancelButton);
+
+    const donateButton = this.createButton(
+      nls.localize('arduino/versionWelcome/donateButton', 'Donate now')
+    );
+    this.addAction(donateButton, this.openDonationPage.bind(this), 'click');
```
per1234, line 70: The dialog should close when the user clicks the "DONATE NOW" button.

per1234 (changes requested):

arduino-ide-extension/src/browser/dialogs/ide-updater/ide-updater-dialog.tsx
```diff
@@ -165,6 +165,37 @@ export class IDEUpdaterDialog extends ReactDialog<UpdateInfo | undefined> {
     goToDownloadPageButton.focus();
   }
 
+  private appendDonateFooter() {
+    const footer = document.createElement('div');
+    footer.classList.add('ide-updater-dialog--footer');
+
+    footer.innerText = nls.localize(
+      'arduino/ide-updater/donateText',
+      'Open source is love, '
```
per1234, line 174: It looks like the space between "Open source is love," and "donate to support us" is lost:

image

per1234 (changes requested):

arduino-ide-extension/src/browser/contributions/check-for-ide-updates.ts
```diff
@@ -64,6 +80,44 @@ export class CheckForIDEUpdates extends Contribution {
...
+    try {
+      const { appVersion } = await this.appService.info();
+      const prevVersion = await this.localStorage.getData<string>(
+        LAST_USED_IDE_VERSION
+      );
+
+      const prevSemVer = new SemVer(prevVersion ?? '');
+      const currSemVer = new SemVer(appVersion ?? '');
+
+      if (prevSemVer.prerelease || currSemVer.prerelease) {
```
per1234, line 114:
```suggestion
      if (
        prevSemVer.prerelease.length > 0 ||
        currSemVer.prerelease.length > 0
      ) {
```

The `prerelease` field will be `[]` when the version is not a prerelease, and `[]` is truthy.

arduino-ide-extension/src/browser/dialogs/ide-updater/ide-updater-dialog.tsx
```diff
@@ -165,6 +165,38 @@ export class IDEUpdaterDialog extends ReactDialog<UpdateInfo | undefined> {
     goToDownloadPageButton.focus();
   }
 
+  private appendDonateFooter() {
+    const footer = document.createElement('div');
+    footer.classList.add('ide-updater-dialog--footer');
+
+    footer.innerText = nls.localize(
+      'arduino/ide-updater/donateText',
+      'Open source is love'
```
per1234, line 174: Would it be possible to do something like this here:

```ts
'Open source is love, {0}'
```

where `{0}` is the placeholder in the template for the donation link?

The reason this would be better is because the current code imposes the grammatical structure `<arduino/ide-updater/donateText>, <arduino/ide-updater/donateLinkIconTitle>`, which won't necessarily be appropriate for all locales. The template approach I describe above allows the translators to adjust the structure of the text as is appropriate for the locale.

The internationalization system supports this template syntax, no problem. For example, you can see it used here:

https://github.com/arduino/arduino-ide/blob/c09b5f718a5a14c420c1649ac5bf117d8fd68d9e/arduino-ide-extension/src/browser/boards/boards-auto-installer.ts#L161-L167

but perhaps it is more challenging in this case where `arduino/ide-updater/donateLinkText` must be linkified.

per1234 (changes requested):

arduino-ide-extension/src/browser/dialogs/ide-updater/ide-updater-dialog.tsx
giacomocusinato: Unfortunately, the internationalization system in Theia only supports replacing primitive values in translation strings; custom elements are not supported. However, I’ve added custom code to allow the anchor element to render at the position defined by the wildcard. This solution currently works only with this specific string format, but we may extend it in the future if this use case becomes more frequent.

per1234 (approved):
Thanks Giacomo!

arduino-ide-extension/src/browser/dialogs/version-welcome-dialog.tsx
```diff
@@ -0,0 +1,104 @@
...
+    const appInfo = await this.appService.info();
+    const { appVersion } = appInfo;
+
+    if (appVersion) {
+      this.titleNode.innerHTML = nls.localize(
+        'arduino/versionWelcome/titleWithVersion',
+        'Welcome to the new Arduino IDE {0}!',
+        appVersion
+      );
+    }
```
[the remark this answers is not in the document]
giacomocusinato: I've changed this specific statement to use `innerText` instead of `innerHTML` which was unnecessary.
This should already pretty be safe but I might leave the sanitizer anyway

martin-eden:
Open source is love!

I wonder who needs that feature? Lot of feature requests from end users?

What is next? Add tracking and ads to IDE?

Issue #2642, opened 2025-02-25
Make the Arduino state update (for extensions) independent of the language server availability


dankeboy36:
### Describe the request

To enhance the reliability of Arduino IDE extensions, IDE needs to ensure that the Arduino state updates are independent of the language server's availability.

As highlighted at https://github.com/dankeboy36/esp-exception-decoder/issues/28#issuecomment-2681800772, the [`compileSummary` of the `ArduinoState`](https://github.com/dankeboy36/vscode-arduino-api/blob/main/docs/interfaces/ArduinoState.md#compilesummary) may be `undefined` due to potential startup failures of the Arduino Language Server. This scenario can lead to issues for extensions developed for the Arduino IDE.

To mitigate this issue, I propose decoupling the `ArduinoState` update process from the language server itself. Given that the compile summary is already accessible, IDE should modify the `compile` command to resolve with a `CompileSummary` rather than `void`. Additionally, the compile command should be relaxed to resolve with `undefined` when the data is partial.

Instead of relying on executed commands to monitor the compile summary, IDE will directly use the resolved compile value to update the state for extensions. Furthermore, the frontend will dispatch the notification that a build has occurred directly from the frontend to the VS Code extension. This should help provide data for extensions reliably and not change the existing IDE behavior.

### Describe the current behavior

IDE backend directly executes the command to update the build path after a verify:

https://github.com/arduino/arduino-ide/blob/6eef09efd8f49827cabb4eabd8cb2ff99ffae09a/arduino-ide-extension/src/node/core-service-impl.ts#L184

IDE updates the `ArduinoState` after the successful command execution:

https://github.com/arduino/arduino-ide/blob/6eef09efd8f49827cabb4eabd8cb2ff99ffae09a/arduino-ide-extension/src/browser/contributions/update-arduino-state.ts#L64-L69

### Arduino IDE version

2.3.4

### Operating system

macOS

### Operating system version

15.3.1

### Issue checklist

- [x] I searched for previous requests in [the issue tracker](https://github.com/arduino/arduino-ide/issues?q=)
- [x] I verified the feature was still missing when using the latest [nightly build](https://www.arduino.cc/en/software#nightly-builds)
- [x] My request contains all necessary details

PR #2643, opened 2025-02-25
fix(plugin): decouple state update from the LS

dankeboy36:
### Motivation

To enhance the reliability of Arduino IDE extensions, the update process for `ArduinoState` has been modified to ensure independence from the language server's availability. This change addresses issues caused by `compileSummary` being `undefined` due to potential startup failures of the Arduino Language Server, as noted in https://github.com/dankeboy36/esp-exception-decoder/issues/28#issuecomment-2681800772.

### Change description

The `compile` command now resolves with a `CompileSummary` rather than `void`, facilitating a more reliable way for extensions to access necessary data. Furthermore, the command has been adjusted to allow resolution with `undefined` when the compiled data is partial.

### Other information

By transitioning to direct usage of the resolved compile value for state updates, the reliance on executed commands for extensions is eliminated. This update also moves the VSIX command execution to the frontend without altering existing IDE behavior.

Closes arduino/arduino-ide#2642

### Reviewer checklist

### DIFF
--- a/arduino-ide-extension/src/node/core-service-impl.ts
+++ b/arduino-ide-extension/src/node/core-service-impl.ts
@@ -176,6 +176,8 @@ export class CoreServiceImpl extends CoreClientAware implements CoreService {
       };
 
       startCompileStream();
+    }).finally(() => {
+      toDisposeOnFinally.dispose();
     });
   }
 
--- a/arduino-ide-extension/src/node/core-service-impl.ts
+++ b/arduino-ide-extension/src/node/core-service-impl.ts
@@ -1,7 +1,6 @@
 import { type ClientReadableStream } from '@grpc/grpc-js';
 import { ApplicationError } from '@theia/core/lib/common/application-error';
 import type { CancellationToken } from '@theia/core/lib/common/cancellation';
-import { CommandService } from '@theia/core/lib/common/command';
 import {
   Disposable,
   DisposableCollection,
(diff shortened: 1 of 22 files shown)


arduino-ide-extension/src/node/core-service-impl.ts
```diff
@@ -164,50 +161,24 @@ export class CoreServiceImpl extends CoreClientAware implements CoreService {
...
+                )}`
+              );
+              resolve(undefined);
+            }
+          });
       };
 
       startCompileStream();
-    }).finally(() => {
-      toDisposeOnFinally.dispose();
```
giacomocusinato, line 172: I think we should keep this? It gets removed in the changes.

giacomocusinato (changes requested):
We need to also add `libsecret-1-dev` as a dependency for Linux tasks to fix the build.
I'm happy to create a new PR by cherry picking your commits or we can update this one; same for me!

arduino-ide-extension/src/node/core-service-impl.ts
dankeboy36: Thank you! Fixed in https://github.com/arduino/arduino-ide/pull/2643/commits/1f1d749cf0e91e5bf96eb23257a2ac546b7fd21a.

dankeboy36:
> I'm happy to create a new PR by cherry picking your commits

I liked your solution (https://github.com/arduino/arduino-ide/pull/2643/commits/b11bde1c473322c86dfde2dd3cec3ebbf92011fa) and cherry-picked it here. Let me know if it's OK for you.

giacomocusinato (approved):
Thanks 🙏

dankeboy36:
Thank you for reviewing the PR multiple times. If it looks good, can you please merge it?

PR #2647, opened 2025-02-27
fix: add missing linux dependencies after ubuntu-latest runner update

giacomocusinato:
### Motivation

Ubuntu 24.04 is the new default version for Ubuntu latest runner https://github.com/actions/runner-images/issues/10636

Ubuntu 24.04 no longer includes the following dependencies:

- `libx11-dev`, `libxkbfile-dev` required by `native-keymap`
- `libsecret-1-dev` required by `keytar` package.

Note that `keytar` is now deprecated and it's a dependency on both IDE2 and Theia https://github.com/eclipse-theia/theia/issues/13593

### Change description

Manually install missing dependencies in CI workflows only for Linux

### Other information

Adopted in favour of https://github.com/arduino/arduino-ide/pull/2649

### Reviewer checklist

- [ ] PR addresses a single concern.
- [ ] The PR has no duplicates (please search among the [Pull Requests](https://github.com/arduino/arduino-ide/pulls) before creating one)
- [ ] PR title and description are properly filled.
- [ ] Docs have been added / updated (for bug fixes / features)

PR #2649, opened 2025-02-27
chore: pin `ubuntu-22.04` runner for linux actions

giacomocusinato:
### Motivation

Ubuntu 24.04 is the new default version for Ubuntu latest runner https://github.com/actions/runner-images/issues/10636

Ubuntu 24.04 no longer includes the `libsecret-1-dev` dependency required by `keytar` on Linux.
`keytar` is now deprecated and it's a dependency on both IDE2 and Theia https://github.com/eclipse-theia/theia/issues/13593

During the transition of getting away from`keytar`, we should use pin runner to `ubuntu-22.04` to ensure `libsecret-1-dev` is present.

### Other information

This might be a better solution then manually installing dependency on each task https://github.com/arduino/arduino-ide/pull/2647

Any suggestion on which solution to adopt is appreciated

### Reviewer checklist

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -89,7 +89,7 @@ env:
           name: Windows_X86-64_zip
     - config:
         name: Linux
-        runs-on: ubuntu-latest
+        runs-on: ubuntu-22.04
         container: |
           {
             \"image\": \"ghcr.io/arduino/arduino-ide/linux:main\"
@@ -140,7 +140,7 @@ env:
 
 jobs:
   run-determination:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-22.04
     outputs:
       result: ${{ steps.determination.outputs.result }}
     permissions: {}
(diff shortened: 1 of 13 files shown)


per1234 (commented):
I think https://github.com/arduino/arduino-ide/pull/2647 is the better solution. The reason being that this solution will stop working when GitHub removes the `ubuntu-22.04` runner, whereas https://github.com/arduino/arduino-ide/pull/2647 is a permanent fix.

The exception would be if we had some indication that the packages will eventually be added to the `ubuntu-24.04` runner, and this is only intended to be a temporary workaround until that time. However, I haven't found anything to indicate that might happen (I did see a report [here](https://github.com/actions/runner-images/issues/10636#issuecomment-2411292886), but no response) and I think the `ubuntu-24.04` runner is fairly mature at this point.

giacomocusinato:
Thanks for the input @per1234. 
I don't have a strong opinion on this, so I agree on keeping `ubuntu-latest` for better long term support.
Eventually if we move away from the `keytar` dependency we can remove the additional manual setup for Linux.

I'm closing this draft in favour of https://github.com/arduino/arduino-ide/pull/2647

PR #2675, opened 2025-04-01
fix: add missing linux dependencies for `merge-channel-files` job

giacomocusinato:
### Motivation

Motivation of this change is the same as https://github.com/arduino/arduino-ide/pull/2647

This specific change has been overlooked in the above PR because `merge-channel-files` runs only when a release is triggered, so the additional error wan't exposed during normal builds. 

### Change description

Add missing linux dependencies for actions that invokes `yarn install`.


per1234 (approved):
Thanks Giacomo!

Issue #2678, opened 2025-04-02
Currently opened sketch loaded when another sketch is opened by command line argument while IDE already running


Gartosz:
### Describe the problem

In addition to opening them via the Arduino IDE GUI, sketches can be opened directly either via file association or by passing their path as an argument to the IDE command line invocation.

If a sketch is opened in this manner while Arduino IDE is already running, the currently opened sketch is opened again in a new window instead of the sketch specified by the command line argument.

### To reproduce

1. Open one sketch.
2. Open different sketch by double clicking on an ino file.
3. Both windows show the first one.

### Expected behavior

I should have multiple sketches in multiple windows without need to use open option in the program.

### Arduino IDE version

2.3.5

### Operating system

Windows

### Operating system version

11

### Additional context

I bisected the regression to 859d29d41af2146537abedb42c9798011b452983 / https://github.com/arduino/arduino-ide/pull/2654 (does not occur when using the build at the previous commit d298b3ffc94008e89066cd999d891e84190da18f).

---

Opening additional sketches via the **File > New Sketch** menu item in the Arduino IDE GUI works correctly.

#### Additional reports

- https://github.com/arduino/arduino-ide/issues/2678#issuecomment-2776370929
- https://github.com/arduino/arduino-ide/issues/2683

### Issue checklist

- [x] I searched for previous reports in [the issue tracker](https://github.com/arduino/arduino-ide/issues?q=)
- [x] I verified the problem still occurs when using the latest [nightly build](https://www.arduino.cc/en/software#nightly-builds)
- [x] My report contains all necessary details

Gartosz:
Also during compile in one window I cannot use top menu in the second one. It closes itself whenever another row is printed in the output. Seems like it impacts other windows too.

nickthesafe:
Confirmed here, the issue is as Gartosz states; have one sketch open and attempt to open further instances of Arduino IDE containing alternate sketches. The program simply duplicates the previously opened instance into the new window and not the file as requested.

kwagnitz:
Ditto on Ubuntu Linux 24.04.
This has been a problem for a very long time.  Makes the IDE confusing to use. 
I don't know why it is so hard to correct.

Gartosz:
> Ditto on Ubuntu Linux 24.04. This has been a problem for a very long time. Makes the IDE confusing to use. I don't know why it is so hard to correct.

Then why haven't you reported that when the issue firstly occured?

kwagnitz:
[...]

PR #2686, opened 2025-04-04
fix: propagate electron params in second instance startup

giacomocusinato:
### Motivation

Resolved https://github.com/arduino/arduino-ide/issues/2678

### Change description

Propagate argv electron params when starting a second instance


davegarthsimpson (approved):

per1234 (approved):
I verified this fixes https://github.com/arduino/arduino-ide/issues/2678.

Thanks Giacomo!

Issue #2696, opened 2025-04-08
Auto update fails on Arduino IDE 2.3.5


giacomocusinato:
### Describe the problem

Arduino IDE uses electron APIs to check and perform auto-updates if a newer version of the application is available.
The auto-update logic is not triggered in IDE 2.3.5 due to a regression in #2654 (https://github.com/arduino/arduino-ide/pull/2654/commits/8053c55ec35b89cea358d456263cf0e8e9986fbd) 

The issue seems to be also present in Theia IDE examples ([here](https://github.com/eclipse-theia/theia/blob/436de3c4571e0000fdcbbf2e94ddfc992f38ccb2/examples/api-samples/src/electron-main/update/sample-updater-main-module.ts#L28))

### To reproduce

- Install Arduino IDE v2.3.5 https://github.com/arduino/arduino-ide/releases/tag/2.3.5
- Open the command palette (on MacOS `Cmd+Shift+P`), look for "Open Settings (UI)" and select it
- Look for the setting "Update Channel" and set it to `nightly` (NOTE: this will set the IDE to receive nightly updates when found)
- Restart the IDE
- A dialog prompting to update to the latest 2.3.6-nightly should appear, but it's not

Note that this also affects stable releases, but switching to nightly channel is the only current way to test this since it's currently the only update available after 2.3.5.

### Expected behavior

Auto-update logic should work and prompt to new any new version if found.

### Arduino IDE version

2.3.5

### Operating system

macOS

### Operating system version

15.3.2

### Additional context

Note that current users that have updated to Arduino IDE 2.3.5 might not receive auto-update for newer versions.

### Issue checklist

- [x] I searched for previous reports in [the issue tracker](https://github.com/arduino/arduino-ide/issues?q=)
- [x] I verified the problem still occurs when using the latest [nightly build](https://www.arduino.cc/en/software#nightly-builds)
- [x] My report contains all necessary details

PR #2697, opened 2025-04-08
fix: use `ElectronConnectionHandler` to connect the ide updater frontend and the electron main process 

giacomocusinato:
### Motivation

Resolves https://github.com/arduino/arduino-ide/issues/2696

### Change description

Use `ElectronConnectionHandler` instead of `ConnectionHandler` to establish communication between the frontend and the electron main process. This is needed specifically for services interacting with electron.

### Other information

The issue seems to appear also on the [Theia IDE example](https://github.com/eclipse-theia/theia/blob/436de3c4571e0000fdcbbf2e94ddfc992f38ccb2/examples/api-samples/src/electron-main/update/sample-updater-main-module.ts#L28). An issue will follow on Theia side.


per1234 (approved):
I verified this fixes https://github.com/arduino/arduino-ide/issues/2696

Thanks Giacomo!

PR #2922, opened 2026-06-08
fix: explicitly get `directories.data` and `directories.user` in separate commands in CLI config init

giacomocusinato:
### Motivation

Arduino IDE uses `arduino-cli config get directories` to get default global system values for `directories.user` and `directories.data` folders if not explicitly set by user. 
This values are needed by IDE to work correctly even if not explicitly returned by  CLI. 

🐛 If only one value is set in global CLI config, `arduino-cli config get directories` only return that value, leaving the other undefined.

### Change description

Get both `directories.data` and `directories.user` fields on separate commands, instead of only fetching its parent (`directories` node).

### Reviewer checklist

### DIFF
--- a/arduino-ide-extension/src/browser/dialogs/settings/settings.ts
+++ b/arduino-ide-extension/src/browser/dialogs/settings/settings.ts
@@ -311,10 +311,13 @@ export class SettingsService {
       );
     }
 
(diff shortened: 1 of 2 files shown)


cmaglie (approved):
Tested on Linux, works as expected 👍🏼

